PR #540, opened 2013-05-22
please add don’t show deck after duel option

wheel-of-fate:
many duelists these days are copy decks what makes real creativity in duel style is no longer exist please add this option "don't show deck after duel"


IceYGO:
This pull request is basically asking to merge ygopro2 into master ...? I guess I can close it...

By the way, not showing the deck after a duel is not possible without breaking the replays.

PR #800, opened 2014-05-22
Update c4239451.lua

Steeldarkeagel:
Fix: Can now be used on monsters with equal attack, and monster effect used on face-down.


salix5:
fixed

PR #802, opened 2014-05-23
Update c8131171.lua

Steeldarkeagel:
Fix: Will not try to activate again after being negated by a card like LADD


salix5:
http://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=4&cid=4481
The effect is completely changed.

PR #821, opened 2014-06-17
fix

sidschingis:
fixed a bug where one was able to tribute opponents monsters when Kaiser Kolloseum is active


salix5:
7aa879d11b2069ea189a9821628893c0f99d5381
fixed
Thank you.

PR #847, opened 2014-07-13
Update c89493368.lua

Steeldarkeagel:
Fix: Will now be allowed to negate one attack if it has used its damage effect that turn.


salix5:
b502f4e396cd025f0863b5caf8b32b1bbd5ad23f
fixed

PR #871, opened 2014-07-25
Update c87602890.lua

Steeldarkeagel:
Fix: Will now work if the monster was Light on the Field but not Light off the Field. Such as by DNA Transplant.


salix5:
fixed

PR #912, opened 2014-08-18
Update c16195942.lua

Steeldarkeagel:
Fix: Will no longer gain attack if the targeted monster is immune to the effect.


salix5:
http://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=13598&keyword=&tag=-1
The current script is correct.

PR #954, opened 2014-10-21
Update c55742055.lua

Steeldarkeagel:
Fix: Is now unable to activate Noble Arms Equip from the hand that are already on the Field under the "Highlander Effect"


salix5:
fixed

PR #1104, opened 2015-02-09
fix

sidschingis:
fixed counting cards returned to the extra deck


salix5:
a306f8bb9a9333df08380654e75802251122924a
fixed

PR #1112, opened 2015-02-13
fix Snapdragon, Envoy of the Evening Twilight, Appointer of the Red Lotus

nekrozar:
Fix this: Banished card doesn't return to the hand to the End Phase.


salix5:
e8666bbabaa2ad7df3c9a305453ec2f2340abea3
fixed
Thank you.

PR #1137, opened 2015-02-27
Shuffle

sidschingis:
added the option for the turn player to manually shuffle his hand in his main phase ( max. 5 times per turn)


salix5:
It seems that
COMMAND_SHUFFLE
btnShuffle
are not well-defined.

sidschingis:
Looks like I forgot to copy over the button definition in the header file.

PR #1145, opened 2015-03-01
attacked count

sidschingis:
if the attack was negated, the attackedCount should not be increased


salix5:
http://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=13111&keyword=&tag=-1
In OCG ruling, the attack is counted even if it is negated.

sidschingis:
that is only for the attack announced count,
for the attacked count it is necessary to enter the Damage Step (see Goblin Attack Force)
http://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=4&cid=5145

PR #1247, opened 2015-04-05
Update c16909657.lua

Steeldarkeagel:
Fix: Will no longer summon tokens or monsters not in the Graveyard that have been destroyed by this card in battle.


salix5:
19ec4643130e40efc89a020cca617759ca1b9f0a
fixed
Thank you.

PR #1248, opened 2015-04-05
Update c34568403.lua

Steeldarkeagel:
Fix: Will no longer summon: Tokens, and Pendulum monsters destroyed and sent to the Extra Deck.


salix5:
19ec4643130e40efc89a020cca617759ca1b9f0a
fixed 
Thank you.

PR #1257, opened 2015-04-12
new effect flag

sidschingis:
new effect flag to ignore the client hint
Hardened Dragon + Earthbound immortals


DailyShana:
I believe it has been fixed here 313897fd87553b30a01283779d8f5127b873fa37

PR #1310, opened 2015-05-02
Startup Parameters

Kelypot:


IceYGO:
Why have you replaced the `strcmp` calls with a manual comparison of the characters?
You are not checking for the terminating null bytes, meaning that `-rsomething` will be accepted.

This condition looks incorrect:
https://github.com/Fluorohydride/ygopro/pull/1310/files#diff-7c1b9672d7d2a0f3473a51edffae0c8dR31
The operator `&&` as a higher precedence than `||`, the dash will only be checked for the option 's'.

PR #1357, opened 2015-05-24
Update c8763963.lua

sidschingis:
should not gain life if target is immune


VanillaSalt:
what you are saying is not be achieved with it

sidschingis:
fixed a typo

salix5:
b258f8e4081b1701342440904cde61bae38d1262
fixed
Thank you.

PR #1380, opened 2015-06-04
Update c55401221.lua

Steeldarkeagel:
Fix: Removed double return.


salix5:
It's my mistake.
Thank you so much!

PR #1456, opened 2015-07-22
fix Remove Brainwashing

DailyShana:
fix https://github.com/Fluorohydride/ygopro/issues/1138 (immunity)

When the effect of Remove Brainwashing stops applying, the controler of Falling Down(equip card)/Tuner's Scheme(continuous target) should take control of opponent's monster again immediately.

### DIFF
--- a/ocgcore/card.cpp
+++ b/ocgcore/card.cpp
@@ -1228,7 +1228,7 @@ void card::reset(uint32 id, uint32 reset_type) {
 			}
 		}
 		if(id & RESET_TURN_SET) {
-			effect* peffect = check_equip_control_effect();
+			effect* peffect = check_control_effect();
 			if(peffect) {
 				effect* new_effect = pduel->new_effect();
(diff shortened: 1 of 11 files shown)


salix5:
想問一下
是刻意改成不使用pcard->reset()嗎？

DailyShana:
希望保留EFFECT_FLAG_SINGLE_RANGE类型的控制权效果，比如调律师的阴谋

DailyShana:
rebased

If the continuous effect target of Tuner's Scheme becomes face down, the effect of Tuner's Scheme will not apply but the control will not return.
http://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=9423&keyword=&tag=-1

For 霊使い/Charmer monsters, I have confirmed that Remove Brainwashing, DNA Transplant(change attribute) and Forbidden Chalice(disable effect of the effect owner) will not reset the control effect.

PR #1475, opened 2015-08-03
Update c59718521.lua

Steeldarkeagel:
Fix: No longer able to Tribute monsters marked as destroyed in battle.


salix5:
c59718521.filter is wrong

Steeldarkeagel:
Thanks salix5 updated the branch,

PR #1486, opened 2015-08-09
fix Latinum, Exarch of Dark World

nekrozar:
Fix this: When Latinum is Special Summoned, you cannnot activate Inferno Reckless Summon.


salix5:
72d9d3cbca690230ac141f898183e95c58a061ca
fixed
Thank you.

PR #1504, opened 2015-08-20
fix

DailyShana:
Even if Pot of Duality has been activated, when Guardian Dreadscythe is sent to grave from field, its effect has to be activated.
http://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=7313&keyword=&tag=-1


salix5:
12514d6130a96714d45a0510380ad9bff8c4d658
已修正
謝謝

PR #1512, opened 2015-08-24
Fix Cyber Shadow Gardna

destdev:
Currently the attack/def gain doesn't work at all (as seen here: http://forum.ygopro.de/index.php/Attachment/1565-bugged-traps-yrp/)

Not sure why the effect type was changed the last time, at least I see this part as a continuous effect in terms of card text and rulings.
Also, not sure why the Code was changed to attack announce instead of be battle target, at least the latest text I see speaks of being targeted for an attack.
This part also seems the cause for the error, changing that alone makes it work again.


salix5:
8d4036d93f09bbc1b7f781586806bb1c6f7150c6
fixed

http://www.db.yugioh-card.com/yugiohdb/?request_locale=ja
or change the language to 日本語

1.
http://www.db.yugioh-card.com/yugiohdb/faq_search.action?ope=5&fid=15062&keyword=&tag=-1
The effect is a trigger effect

2.
サイバー・シャドー・ガードナー
相手のメインフェイズ時に発動できる。このカードは発動後モンスターカード（機械族・地・星４・攻／守？）となり、自分のモンスターカードゾーンに特殊召喚する。このカードを攻撃対象とした相手モンスターの攻撃宣言時、このカードの攻撃力・守備力は相手攻撃モンスターと同じ数値になる。このカードは相手のエンドフェイズ時に魔法＆罠カードゾーンにセットされる。このカードは罠カードとしても扱う。

Its timing is 攻撃宣言時/EVENT_ATTACK_ANNOUNCE, which means it cannot be activated if the battle is rolled back.
This is different from EVENT_BE_BATTLE_TARGET

destdev:
thanks for the explanation

PR #1526, opened 2015-09-01
fix Starduston

nekrozar:
Fix this: If your opponent controls more monsters than you do, Starduston is destroyed.


salix5:
It's my mistake.
Thank you very much!

PR #1528, opened 2015-09-01
display number of pendulum in extra deck

DailyShana:
also send Fusion/Synchro/Xyz monsters to the bottom of extra deck, making pendulum cards always on the top of extra deck


DailyShana:
snapshot:
image

PR #1534, opened 2015-09-05
fix Darklight

nekrozar:
Fix this: If monsters are not Special Summoned this turn is returned to the field, it is destroyed by "Darklight" effect.


salix5:
52b02efd827b82e2fef378e29448709c0eaec5de
fixed
Thank you!

PR #1535, opened 2015-09-05
Immune effect of Qli monsters

DailyShana:
In a more natural approach


VanillaSalt:
if peffect->reason_effect is reset by the previous turn, the effect is erased in field::process_turn

DailyShana:
Oh, sorry, this idea doesn't work

PR #1549, opened 2015-09-12
do not adjust self destroy set in the process of move_to_field

DailyShana:
when you get control of a Burning Abyss monster by the effect of No.11, that monster is sent to grave and then at the timing to activate its effect the program crash. The reason is in the process of get_control, when it is self destroy, the controler of No.11 still get control of that monster in grave.
I think in the process of move_to_field, adjusting self destroy set should not be raised, it has caused many curious problem.
According to http://yugioh-wiki.net/index.php?%A3%D3%A3%E9%A3%EE
Ｑ：フィールド魔法が存在していない状態で特殊召喚した場合その特殊召喚に対して《激流葬》を発動することは可能ですか？
Ａ：いいえ、できません。(11/06/08)

Ｑ：上記の例でフィールド上に他のモンスターが存在する場合は、《激流葬》を発動できますか？
Ａ：自分または相手は《激流葬》を発動できます。(12/07/06)
EVENT_CONTROL_CHANGED should also be raised even it has been self destroyed.

### DIFF
--- a/ocgcore/operations.cpp
+++ b/ocgcore/operations.cpp
@@ -1021,19 +1021,13 @@ int32 field::control_adjust(uint16 step) {
 		card_set* adjust_set = (card_set*)core.units.begin()->ptarget;
 		delete adjust_set;
 		core.control_adjust_set[0].insert(core.control_adjust_set[1].begin(), core.control_adjust_set[1].end());
-		for(auto cit = core.control_adjust_set[0].begin(); cit != core.control_adjust_set[0].end(); ) {
-			card* pcard = *cit++;
-			if(!(pcard->current.location & LOCATION_ONFIELD)) {
-				core.control_adjust_set[0].erase(pcard);
-				continue;
-			}
-			pcard->filter_disable_related_cards();
-			if(pcard->unique_code)
-				add_unique_card(pcard);
-			raise_single_event(pcard, 0, EVENT_CONTROL_CHANGED, 0, REASON_RULE, 0, 0, 0);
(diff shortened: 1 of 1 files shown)


salix5:
可否上傳錄像
目前測試用No.11巨眼抓彼岸之後會正常破壞

DailyShana:
前提是那个被破坏的彼岸怪兽的遗言效果可以正常发动，例如夺取彼岸の悪鬼 グラバースニッチ(20758643)，对方卡组有其他的彼岸怪兽

salix5:
e25878de68817f88fee57c946c2fce272c114009
目前先嘗試修改move_to_field()
直接奪取控制權的效果通常是
duel_get_control()

get_control()
step 0:
move_to_field(pcard, playerid, playerid, LOCATION_MZONE, pcard->current.position);
step 1:
set_control(pcard, playerid, reset_phase, reset_count);

void move_to_field(card* target, uint32 move_player, uint32 playerid, uint32 destination, uint32 positions, uint32 enable = FALSE, uint32 ret = FALSE, uint32 is_equip = FALSE);
target->operation_param = (move_player << 24) + (playerid << 16) + (destination << 8) + positions;
add_process(PROCESSOR_MOVETOFIELD, 0, 0, (group*)target, enable, ret + (is_equip << 8));

int32 field::move_to_field(uint16 step, card * target, uint32 enable, uint32 ret, uint32 is_equip)

DailyShana:
我就是这样改的呀，只是把control adjust的step 4也改回去了

PR #1598, opened 2015-10-27
Fix RELOAD_FIELD message handling

IceYGO:
The RELOAD_FIELD message handling is inconsistent with the core.
See: https://github.com/Fluorohydride/ygopro-core/blob/d1e09e5a63c79f21bccbfee5c83f5fd636fac398/field.cpp#L128

Also, we need a second loop in order to add the hand cards to prevent an incorrect positioning since the position of a card depends on the hand count.


IceYGO:
My bad, used the branch master for the pull request. Will one another one with the correct branch.

PR #1599, opened 2015-10-27
Fix RELOAD_FIELD message handling

IceYGO:
The RELOAD_FIELD message handling is inconsistent with the core.
See: https://github.com/Fluorohydride/ygopro-core/blob/d1e09e5a63c79f21bccbfee5c83f5fd636fac398/field.cpp#L128

Also, we need a second loop in order to add the hand cards to prevent an incorrect positioning since the position of a card depends on the hand count.


Zayelion:
@DailyShana 
@salix5 
@VanillaSalt 

Please review, needed for other YGOPro variants to function.

salix5:
Sorry for the delay.
Thank you very much!

Zayelion:
Thank you! Very much appreciated we can all rebuild our servers now!

DailyShana:
Although 3 years passed, I think we can call `mainGame->dField.RefreshAllCards()` after adding cards to avoid incorrect position in hand, just like the treatment in single mode.

PR #1600, opened 2015-10-27
Allow spaces in strings.conf and some fields in system.conf

IceYGO:
Latin languages like English use spaces to separate words in their sentences. However, the current configuration files are limited to one word (because of `%s`).

This commit allows spaces to be used in `strings.conf` and `system.conf`, so we can properly translate the game.

### DIFF
--- a/gframe/data_manager.cpp
+++ b/gframe/data_manager.cpp
@@ -85,19 +85,19 @@ bool DataManager::LoadStrings(const char* file) {
 			continue;
 		sscanf(linebuf, "!%s", strbuf);
 		if(!strcmp(strbuf, "system")) {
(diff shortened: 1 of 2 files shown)


Zayelion:
@Fluorohydride
@VanillaSalt
@salix5 

Could you guys please pull this in so we can fix all the YGOPro's that arent in asia.... that would be very kind.

IceYGO:
:arrow_up:

salix5:
Thank you so much!

IceYGO:
Thanks you for the merge and improvement commit!

Zayelion:
@salix5 OMG ***THANK YOU!**

PR #1612, opened 2015-11-14
fix

DailyShana:
When choosing one ignition effect from two (eg Lavalval Chain) and then press Enter BP button, the Option window should be hidden.


DailyShana:
to avoid the following situation: you can still choose one of the ignition effect to activate in Battle Phase, and if you do, it will cause an error (MSG_RETRY).
image

PR #1632, opened 2015-12-20
increase buffer

sidschingis:
increase buffer to allow for more than 40 main deck cards in duel puzzle


IceYGO:
:+1:

A single card requires 88 bytes. A 60-cards deck requires 5280 bytes.
This commit increases the buffer from 4096 to 8192 bytes.

sidschingis:
A card uses 92 bytes and the 4096 bytes buffer seemed to be pretty arbitrary in the first place, since you only need 3680 bytes for 40 cards.

IceYGO:
My bad, I forgot one variable, I edited the calculation.

Yes, I definitely agree with 8192, it seems enough for the foreseeable future, even if we add some variables. I was pointing out that 4096 is definitely not enough.

PR #1641, opened 2015-12-31
Use LocalPlayer() to retrieve the correct player index in MSG_RELOAD_FIELD

IceYGO:
Every message uses `LocalPlayer()` to retrieve the player index, in case the game was swapped.

`MSG_RELOAD_FIELD` is the only message that does not do this, causing problems whenever the host is not the starting player.


IceYGO:
Thanks!

PR #1660, opened 2016-01-25
cover2, bg_menu, bg_deck

DailyShana:
cover2.jpg, bg_menu.jpg, bg_deck.jpg are optional and can be deleted safaly.


IceYGO:
The default background is always drawn, even when another background is drawn over it.
When we are in the menu or the deck editor, it means one useless `draw2DImage` call per frame.

Well, I don't think drawing one more image is gonna hurt the GPU and beside that everything looks good, so it's a +1 for me. :+1:

adrianojn:
+1
works ok

Issue #1634, opened 2015-12-21
Sounds and Music


Zayelion:
## Music
- https://github.com/sidschingis/Dev_ygopro/commit/dca13aeab3ff7df536b8ea313652ecdfce1508b7
- https://github.com/sidschingis/Dev_ygopro/commit/3dc2c474de67c75cf8e8f7148ef4c7caacdcb0b9
## Sounds
- https://github.com/Buttys/ygopro/commit/7f86c719598c2e3d04bd95968c4bcd1db0898daf
- General massive change worth reviewing [ https://github.com/Buttys/ygopro/commit/3f98f645bb874e4b306b727d2a05eea5ca36bfc6 ]
## Controls
- https://github.com/Buttys/ygopro/commit/a8a30be6779238a70cd58df84f7155c19d2c96b9
- https://github.com/Buttys/ygopro/commit/ad65b336fe39d181c9beff29b85be347a8f0f752

Zayelion:
I remember there being some contention on the use of the `PlaySound` function because it was proprietary or something of that nature?

Zayelion:
Update : http://www.ambiera.com/irrklang/license.html
@Fluorohydride
@VanillaSalt
@salix5 

Are you guys ok with this license?

IceYGO:
Personally, I am against irrKlang because they are not open-source. All the libraries used by YGOPro are open-source and should remain so.

Zayelion:
I found a way of doing it using windows' built in headers but the sounds dont overlap and its ofcourse windows specific which Im against.

DailyShana:
It seems has been done?

mercury233:
it will be better to add another sound engine support

PR #1662, opened 2016-01-25
Sound

DailyShana:


Zayelion:
1.) This works and covers use case, its the way its done in two of the variants.
2.) irrklang has some licencing issues with the principles of being open source. http://www.ambiera.com/irrklang/faq.html#opensource. If someone wants to use YGOPro commercially in the future (say Konami, or a group/company that has licenced YuGiOh properly), @Fluorohydride declared that they have to share the source for the gframe used. https://github.com/Fluorohydride/ygopro/blob/master/LICENSE ygopro-core is MIT so it doesnt have to be shared, https://github.com/Fluorohydride/ygopro-core/blob/master/LICENSE so server side systems are protected... so I dont know if these licences are comparable.
3.) I feel for principle's sake we need to find a difference sound engine and keep this commit for a record.

IceYGO:
As I said in #1634, I do not like irrKlang since it is not using an open-source license.
I'm not against including it but only if we can also compile without it.

It's sad to see how few free audio libraries there are. :disappointed:
Maybe we could take a look at http://www.portaudio.com/, http://www.sfml-dev.org/ or https://www.openal.org/.

Zayelion:
I prefer SFML, but @Fluorohydride why wasnt sound originally included?

zh99998:
+1 for this request.

about licensing thing.
if someone (say Konami, or a group/company that has licensed YuGiOh properly), do you say Konami or company who can buy license from Konami, can't afford a audio library? this worry is stupid.

fervi:
We can use music with 0AD. License is ok.
Ammon-Ra - fight
Honor_Bound - if winning
Taiko 1 / Taiko 2 - if losing

Victory Track - if win
Dried Tears - if lose

Issue #1658, opened 2016-01-18
Case Sensitivity : Cards that Require Name Declaration Broken Fail Seach


Zayelion:
In cases of cards like Archfiend Oath, in English atleast the case sensitivity is bugged.

If capitalization is off:
![http://puu.sh/mmcZA/a730533dbb.jpg](http://puu.sh/mmcZA/a730533dbb.jpg)

If spelled and cased correctly
![http://puu.sh/mzL5h/105022ae2a.jpg](http://puu.sh/mzL5h/105022ae2a.jpg)

a fix can be found here :
https://github.com/Buttys/ygopro/commit/e117936e24ebcd091a68d1c5f6b492bd7df4d970#diff-249012b053d63d10e4cc07838658c267

Zayelion:
From a user:
> When I activate mind crush, the list and search bars are extremely limited.

From a user :
> I believe this might be a problem with other card declaration effects .. i couldn't declare any of the sylvan monsters, tried with capital letters but could only declare the spells.. couldn't find lonefire either
![https://i.gyazo.com/9022f799deab2ea06cfe31edfce7ff0d.jpg](https://i.gyazo.com/9022f799deab2ea06cfe31edfce7ff0d.jpg)

Zayelion:
https://github.com/Fluorohydride/ygopro/pull/1663

salix5:
My opinion:
Case sensitive is not a bug, since the card names in the official database do have uppercase and lowercase letters, and  the names are case sensitive.
I believe that XYZ-Dragon Cannon is not an "Xyz" card without further mention is an example.

PR #1663, opened 2016-01-25
Search the cards while typing in the deck editor and ignore the case

Zayelion:
fix for https://github.com/Fluorohydride/ygopro/issues/1658

### DIFF
--- a/gframe/deck_con.cpp
+++ b/gframe/deck_con.cpp
@@ -86,101 +86,7 @@ bool DeckBuilder::OnEvent(const irr::SEvent& event) {
 				break;
 			}
 			case BUTTON_START_FILTER: {
-				filter_type = mainGame->cbCardType->getSelected();
-				filter_type2 = mainGame->cbCardType2->getItemData(mainGame->cbCardType2->getSelected());
-				filter_lm = mainGame->cbLimit->getSelected();
-				if(filter_type > 1) {
-					FilterCards();
-					break;
-				}
-				filter_attrib = mainGame->cbAttribute->getItemData(mainGame->cbAttribute->getSelected());
-				filter_race = mainGame->cbRace->getItemData(mainGame->cbRace->getSelected());
-				const wchar_t* pstr = mainGame->ebAttack->getText();
-				if(*pstr == 0) filter_atktype = 0;
-				else {
-					if(*pstr == L'=') {
-						filter_atktype = 1;
-						filter_atk = BufferIO::GetVal(pstr + 1);
-					} else if(*pstr >= L'0' && *pstr <= L'9') {
-						filter_atktype = 1;
-						filter_atk = BufferIO::GetVal(pstr);
-					} else if(*pstr == L'>') {
-						if(*(pstr + 1) == L'=') {
-							filter_atktype = 2;
-							filter_atk = BufferIO::GetVal(pstr + 2);
-						} else {
-							filter_atktype = 3;
-							filter_atk = BufferIO::GetVal(pstr + 1);
-						}
-					} else if(*pstr == L'<') {
-						if(*(pstr + 1) == L'=') {
-							filter_atktype = 4;
-							filter_atk = BufferIO::GetVal(pstr + 2);
-						} else {
-							filter_atktype = 5;
-							filter_atk = BufferIO::GetVal(pstr + 1);
-						}
-					} else if(*pstr == L'?') {
-						filter_atktype = 6;
-					} else filter_atktype = 0;
-				}
-				pstr = mainGame->ebDefence->getText();
-				if(*pstr == 0) filter_deftype = 0;
-				else {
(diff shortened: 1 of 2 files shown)


IceYGO:
I cannot very well review my own commit, so if anyone else want to take a look at it...

I am pretty sure this commit only affect the deck editor and not the in-game dialog.
It is probably just a line to change in order to make it work though.

Zayelion:
True, given the complexity and lack of a test framework someone needs to compile and test this.

adrianojn:
:+1:  LGTM

Zayelion:
[LGTM](http://lgtm.in/i/orJ1)

[:+1:](http://lgtm.in/u/orJ1) [:-1:](http://lgtm.in/r/orJ1)

salix5:
My opinion:
Case sensitive is not a bug, since the card names in the official database do have uppercase and lowercase letters, and the names are case sensitive.
I believe that XYZ-Dragon Cannon is not an "Xyz" card without further note in wikia is an example.

Zayelion:
That logic is faulty. **It fundamentally creates difficultly using the program** if the user is not familiar with the exact capitalization of the cards. It has nothing to do with archetypes, there is a separate search filter for that.

DailyShana:
We can introduce an option called 'latin=1' to enable case-free searching
and fixing the alignment for English.

Zayelion:
> We can introduce an option called 'latin=1' to enable case-free searching
and fixing the alignment for English.

and the `r` key changing the font so it is illegible.

adrianojn:
@salix5 please read about [Letter case](https://en.wikipedia.org/wiki/Letter_case)

In particular:

> If an alphabet has letter case, all or nearly all letters have both forms. Paired forms are considered **variants of the same letter**: they have the **same name and pronunciation** and will be **treated identically when sorting** in alphabetical order.

So yes, this is a bug for *all western languages*.

This is an example of a bug in Ygopro caused by cultural differences (only western languages have letter case). Others examples are:
* Ygopro does not support left-to right languages
* [Word wrapping](https://en.wikipedia.org/wiki/Line_wrap_and_word_wrap#Word_boundaries.2C_hyphenation.2C_and_hard_spaces) is wrong for all languages expect CJK

Zayelion:
- [ General guidelines for building websites](http://www.usabilityfirst.com/glossary/case-sensitivity/)
- [Statement from the United States Government ](http://guidelines.usability.gov/guidelines/129) this is basically law. In the United States for good [reason](http://guidelines.usability.gov/guidelines/136). 
- [quote/use case example from the leading expert on internet usability, Jakob Nielsen](https://books.google.com/books?id=DBOowF7LqIQC&pg=PA148&lpg=PA148&dq=case+sensitivity+in+search+usability&source=bl&ots=Bk5aXRLSzQ&sig=MmRxMw9Ufeu9NpZA-GEJEE7wm2k&hl=en&sa=X&ved=0ahUKEwitzbnthdrLAhXJdj4KHcIGADYQ6AEIMjAE#v=onepage&q=case%20sensitivity%20in%20search%20usability&f=false)
- [A statistical analysis of what people type in](https://searchenginewatch.com/sew/news/2048984/search-engines-and-capitalization)
- [a break down](http://architectingusability.com/2012/07/23/designing-search-systems/)

This issue isnt unique either, this is a generic programmer oversight because its *harder* to do.
- [Irony](https://github.com/gitlabhq/gitlabhq/issues/3108) Gitlab
- [More Irony](https://jira.atlassian.com/browse/JRA-24907) Note the mocking user

Its hard to quantify why this *"feels wrong to the user"*, but the point is, a *user* interface should be user focused.

salix5:
About letter case:
I agree that the query text should be case-insensitive now, since I didn't notice that many databases(including the official card database) do take this approach.

About NormalizeChar:
My opinion is that we should not put any limit on the query text except for the preserved chars(@ for setname, $ for card name). There are cards has signs like #. / in their names, so I think that converting the letter case is enough.

About EGET_EDITBOX_CHANGED:
This feature may be convenient in alphabetical language like English, but it is somewhat unnecessary in languages like Chinese. For example, typing 機殼(Qli) will search for 機 first, which will find 500+ irrelevant cards since this character appears in 機械族(Machine type), and many other setnames/nouns have similar problems.

DeckBuilder  may involve some language-specific details, and I don't know much about this part.
Since DeckBuilder is a rather independant module, maybe you can put some of these features in other language-specific branches?

DailyShana:
I agree.
In fact, it is more of the issue of irrlicht than ygopro itself.

mercury233:
will add a switch for auto search in system.conf

mercury233:
https://github.com/Fluorohydride/ygopro/commit/3d75697d7015f0df1a6dd249c87dcb65f4b724c7 I agree with salix5 on NormalizeChar.

PR #1674, opened 2016-02-06
update lflist

mercury233:
2016.2 TCG
http://www.yugioh-card.com/en/gameplay/errata/FL_List_Feb2016.html


DailyShana:
It is not a formal list

https://yugiohblog.konami.com/articles/?p=7596
> The full, applies-everywhere Forbidden & Limited Cards List will be updated in a month or two in its usual manner, and some (perhaps all) of these Adjusted List restrictions will likely get rolled into that. (And some of you are going to ask, “Are we going to get an Adjusted List from now on, half-way between F&L Lists?” And the answer is “Probably not, this is an extreme case. But we’ll see how things go.” If you do attend a YCS Event, be sure and come up to let us know your thoughts on this!)

PR #1691, opened 2016-03-06
update strings.conf

nekrozar:
- Add Magician's japanese setname and Galaxy-Eyes Tachyon Dragon's japanese setname
- ~~Remove japanese setname ruby~~
- Fix Majestic's japanese setname and Ojama's japanese setname

### DIFF
--- a/strings.conf
+++ b/strings.conf
@@ -452,8 +452,8 @@
 !setname 0x1 正义盟军	A・O・J
 !setname 0x2 次世代	ジェネクス
 !setname 0x1002 真次世代	レアル·ジェネクス
-!setname 0x2002 盟军·次世代	A·ジェネクス
-!setname 0x3 N/A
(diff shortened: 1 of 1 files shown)


DailyShana:
I think in the following cases, ruby should be recorded:
1. the base of a setname appears with different ruby.
eg. 白竜の聖騎士 is not 聖騎士(せいきし) card.
2. the base is consist of one or two English letters and the ruby has actual meaning

nekrozar:
Add ruby. (Sorry, I thought ruby doesn't need...)
I think that it is better to add English setname.

DailyShana:
English setname should be used in English version
Japanese setnames here is only treated as reference because whether a card belong to a series depends on its Japanese name.

nekrozar:
I understand.

Issue #1609, opened 2015-11-07
How to compile ygopro2


Zayelion:
can I get **exact** steps on how to setup an ygopro2 branch development environment?

salix5:
I don't know much about ygopro2 branch.
Maybe you can ask @Fluorohydride .

Zayelion:
@Fluorohydride

Fluorohydride:
the project need following libraries:
libevent, lua5.2, sqlite3, opengl, glew, glfw3, glm, freetype2
on linux and osx, just install these libs by package manager, and then generate the project file using cmake.
on windows, i think you have to build and link these libs manually.

Zayelion:
I'm on windows using Visual Studio, please walk me through it like I am a complete moron ^_^; the sub-teams around YGOPro use compiling it as a hazing technique for joining up. Its not eactly staightforward for people that are not programmers, and I've found that many people use their passion for YuGiOh to become programmers.

I understand what you said, and it answers my question fully, I am just asking for clarification for the sake of others.

Moonlightsorrow:
How do i exactly compiler ygopro2.

meijieru:
@Fluorohydride lacking CMakeLists.txt at directory `netc` and `ocgcore`

blakat360:
I am having the same issue. I am on Arch Linux. I can't seem to get premake4, so I tried premake5 but ran into issues as the install command couldn't open ocgcore.

I then tried cmake. It complained about the add_subdirectory(ocgcore) call as it contained no CmakeLists.txt file. 

I then subbed this out for an include_directories call, and cmake terminated fine. However, when I call make, I get the following errors.

```bash
n file included from /home/sigkill/src/ygopro/gframe/deck_con.cpp:1:
/home/sigkill/src/ygopro/gframe/config.h:79:10: fatal error: ../ocgcore/ocgapi.h: No such file or directory
   79 | #include "../ocgcore/ocgapi.h"
      |          ^~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [gframe/CMakeFiles/ygopro.dir/build.make:161: gframe/CMakeFiles/ygopro.dir/deck_con.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
In file included from /home/sigkill/src/ygopro/gframe/client_field.h:4,
                 from /home/sigkill/src/ygopro/gframe/client_field.cpp:2:
[...]

PR #1730, opened 2016-04-21
Ygopro2

ronaldoguedess:
How do I test these scripts? I can not find the executable file and was not informed that the engine for me to open the project and see!


mercury233:
Sorry, but you should use issue to ask question.

For your question, see https://github.com/Fluorohydride/ygopro/issues/1609

PR #1734, opened 2016-04-27
counter

mercury233:
https://github.com/Fluorohydride/ygopro-scripts/pull/393


destdev:
hu?
ehm... why didn't they update that in the update yesterday... kk, sorry for missing that

PR #1767, opened 2016-05-30
add chain buttons

mercury233:
Add switch for ignore_chain and always_chain, add chain_when_avail (ask chain at any timing available)

image


Zayelion:
This is great but we cant deploy this due to our systems being on different branches T.T

PR #1826, opened 2016-08-27
Update game.cpp

Tianchenglipu:
If the program doesn't know what the setcode is, just hide it rather than showing to user "???".
@DailyShana


DailyShana:
It indicates player to update strings.conf

PR #1858, opened 2016-10-26
add show OCG-only and TCG-only info in card info

mercury233:
image
image


DailyShana:
f384c7f4db716c94973f879e6af2303cf2aa7c66

Issue #1878, opened 2016-12-02
Changing ot


Chen-ygopro:
I feel that the current ot system is incorrect and creates more work than needed. I feel we should use the power of two for ot's. Also I feel having only 4 possible OTs is quite limiting.
@Zayelion @lllyasviel 

> - 1 = OCG
> - 2 = TCG
> - 4 = Anime
> - 8 = Korean
> - 16 = Beta
> - 32 = Custom
> - 64 - VG

If you want to host a OCG+TCG duel, you simply use 1 | 2 = 3.
If you only want TCG+Anime, you can use 2 | 4 = 6.
If you want to use OCG, TCG and Beta cards, you use 1 | 2 | 16 = 19.
... you get the idea.

我觉得现在的 OT 系统不对, 
我想我们应该使用两个的力量 
只能有四个 OT 有点太限制了。如果要主办 OCG + TCG 决斗  1 | 2 = 3. 
所以等等

Zayelion:
I agree with this implementation.

Hello23-Ygopro:
@Chen-ygopro Thank you , it would be nice if they can make it official

Btw, what about Manga ? Despite popular belief and Ygopro Percy's incorrect naming of manga cards as "(anime)" , anime & manga cards aren't the same , or do u prefer them part of the same OT ?

Otherwise , may i suggest :
- 1 = OCG
- 2 = TCG
-  4 = Anime
- **8 = Manga**
- 16 = Korean
- 32 = Beta
- 64 = Custom
- 128 - VG

I assume VG is video game so video game-exclusives or video game cards with different effects of anime & real life cards

I wanted to suggest adding Chinese there too since they got their own banlist some time ago , but it looks like u only add an OT if there are exclusives involved, and the chinese card game so far has no exclusives yet

mercury233:
if you want to expand the ot system, you should use config file instead of write constants in code.
the code in this repo will keep up to official cards so i guess no extra ot mark will be added.

DailyShana:
We don't consider unofficial cards here so we will not add new ot types.

Zayelion:
It should not be hard coded in the first place.

Chen-ygopro:
What does you considering cards have to do with adding ots? Adding the extra ot's don't force you to use them.

Chen-ygopro:
All of this is the result of bad design choices.

Zayelion:
Bad design can be corrected.

Chen-ygopro:
[...]

PR #1879, opened 2016-12-02
Chen ygopro ot patch 1

Chen-ygopro:
#1878


Zayelion:
There is a OT check somewhere else in the codebase that you missed. Its in server side ban-list validation.

PR #1916, opened 2017-01-20
add hint for select place

mercury233:
image


Hello23-Ygopro:
@mercury233 Alot of English Ygopros hasnt translated this yet
So is this a accurate translation would you say ?
1. `!system 570 Select a Zone to become unusable.`

2. Im not sure how to translate `!system 569 请选择卡片的位置` ..
*Google Translate* says "Select the location of the card." , but i dont know

3. Lastly , could you also tell me where Ygopro uses this string ? 
`!system 66 持续公开`

Thank You !

mercury233:
570 is ok
569 is ok too, it will appear if you turn off auto select location in settings
66 mean "keep public", but i don't know where it is used.

PR #1945, opened 2017-03-21
update lflist.conf

nekrozar:
Next Forbidden & Limited card lists
http://www.yugioh-card.com/japan/event/shop_duel/limit_regulation.php?list=201704


nekrozar:
TCG's Next Forbidden & Limited card lists
http://www.yugioh-card.com/en/limited/index.html

PR #1949, opened 2017-03-26
Added DrawLinkedZones()

edo9300:
This will draw a yellow pulsing recatngle, like the white one, when a zone is hovered, in the monster zones linked to a link monster
image

### DIFF
--- a/gframe/drawing.cpp
+++ b/gframe/drawing.cpp
@@ -198,28 +198,28 @@ void Game::DrawLinkedZones(ClientCard* pcard) {
 	matManager.mSelField.AmbientColor = 0xff0261a2;
 	driver->setMaterial(matManager.mSelField);
 	if (dField.hovered_sequence<5) {
-		if (mark & 0x10 && dField.hovered_sequence>0)
+		if (mark & LINK_MARKER_LEFT && dField.hovered_sequence>0)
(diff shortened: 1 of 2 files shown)


DailyShana:
maybe blue is better?
consistent with the color of Link monsters

edo9300:
@DailyShana i changed the color code with this blue 0261a2
image

mercury233:
I prefer the yellow version.

mercury233:
image
image
It don't work on opponent's zone for me.

gframe/drawing.cpp
```diff
@@ -188,6 +193,36 @@ void Game::DrawBackGround() {
 		driver->drawVertexPrimitiveList(vertex, 4, matManager.iRectangle, 2);
 	}
 }
+void Game::DrawLinkedZones(ClientCard* pcard) {
+	int mark = pcard->link_marker;
+	matManager.mSelField.AmbientColor = 0xff0261a2;
+	driver->setMaterial(matManager.mSelField);
+	if (dField.hovered_sequence<5) {
+		if (mark & 0x10 && dField.hovered_sequence>0)
```
VanillaSalt: LINK_MARKER_LEFT = 0010 (octal)  not 0x10 (hexadecimal)

edo9300:
@VanillaSalt so they are in octal? Sorry I misunderstood
@mercury233 I too think that yellow looks better

Zayelion:
> maybe blue is better?

Stop arguing about nonsense both of you. Implement the theme system that was discussed earlier so this is a non-issue.

PR #1957, opened 2017-03-31
update Field texture

nekrozar:


Hello23-Ygopro:
Will there be an option to have different fields show up for different rules ? for example ,

if master rules 1-2 was checked , then field.png is used
master rules 3 will use field2.png
current master rules use a field3.png

nekrozar:
https://github.com/Fluorohydride/ygopro/blob/master/gframe/image_manager.cpp
@Hello23-Ygopro YGOPRO don't use field.png.

Zayelion:
@nekrozar thats not what he means. He is speaking about obsolete ruling changes.

nekrozar:
https://github.com/Fluorohydride/ygopro/pull/1957/commits/641859adea087160b098124b2c04d4fc8863a507 Is this what you mean?

Hello23-Ygopro:
> 641859a Is this what you mean?

Yes , thank you
i was hoping that the pendulum field does not get overwridden , but make field3.png for the new one

Unfortunately i think it may be wrong , heres how the official link field looks : 
link_image
As you can see the pendulum zones have been removed, and it looks alot more like field.png now, except it has extra monster zones
# 

> YGOPRO don't use field.png.

True
But with the new rules option : https://github.com/Fluorohydride/ygopro/blob/master/strings.conf#L282
i was hoping that Ygopro can play all different fields : field.png , field2.png & field3.png

So if you select "Master Rules 1 or 2" before the duel , then Ygopro loads field.png
if "Master Rules 3" is selected , then Ygopro loads field2.png
and with the new current rules it loads the new field3.png

Sorry my english is not well , does that make sense ?

edo9300:
@Hello23-Ygopro with this https://github.com/Fluorohydride/ygopro/pull/1974 , now the game will do that, if you're playing under mr4, it will load the field with emzones and no pzones, while if you're playing under mr3 or previous it will load the old field

Hello23-Ygopro:
@edo9300 Will u be adding Mr.1 and Mr.2 fields too ? I think its the one that looks like Mr.4 field , but without the extra monster zones

Thank you !! 👍👍

edo9300:
@Hello23-Ygopro I tought to add that too, but then I realized that it would have been useless since you can use pendulums in both mr1 and 2

Hello23-Ygopro:
@edo9300 Ok , i understand , cus even though pendulums didnt exist in Mr.1 and Mr.2 , it wont stop people from using them with those rules

So its safer to have zones for them when people do decide to use them

Thank you

PR #1960, opened 2017-04-05
Added Link Markers Filter

edo9300:
This is how will it look like
image
image

The link marker string is the 1374, but i don't know how is Link Marker said in chinese, so if someone could tell me i'll change the strings too


mercury233:
Link Marker = 连接标记

-----
image
another two layouts, which is better?

Hello23-Ygopro:
I cant say which is better for a Ygopro , but 1 and 2 is definitely neater then 3 imo , and i like that the marker buttons are in the same section as the scales and level/ranks
Personally i like 1 the most

PR #1969, opened 2017-04-14
Fix crashes in MacOS

zhykzhykzhyk:
* Prevent GL method being called from other threads

### DIFF
--- a/gframe/duelclient.cpp
+++ b/gframe/duelclient.cpp
@@ -69,11 +69,11 @@ void DuelClient::ConnectTimeout(evutil_socket_t fd, short events, void* arg) {
 		mainGame->btnCreateHost->setEnabled(true);
 		mainGame->btnJoinHost->setEnabled(true);
 		mainGame->btnJoinCancel->setEnabled(true);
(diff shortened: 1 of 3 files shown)


mercury233:
Personally I prefer not to introduce new libraries, so I wonder if it is possible to fix this in another way.

zhykzhykzhyk:
New libraries? You mean STL?

mercury233:
Yes, and I said it is just personally preference of mine `_(:з」∠)_`.

zhykzhykzhyk:
Do you mean that both the queue and the type erased function need to be implemented again in this project even it is already there in the STL, or there is a technic you are trying to avoid?

mercury233:
I mean we should use new library only if necessary.

Zayelion:
If its in `std` isn't it already included? By don't add new libs dont you mean stuff like `irrKlang` and `sfml`?

xiliu98:
If stl library is forbidden to use, then what stl-like library else can we use?

zhykzhykzhyk:
Close in favor of #2025

PR #1971, opened 2017-04-15
Fix inconsistent issue in deck construction and side changing screen

zhykzhykzhyk:
Update hovered state after every mouse event to keep track with hovered card when layout changed.
Also fixed a crashing issue: when user do right click in a row, the program tries to remove a card that has already been removed.

### DIFF
--- a/gframe/deck_con.cpp
+++ b/gframe/deck_con.cpp
@@ -688,83 +688,6 @@ bool DeckBuilder::OnEvent(const irr::SEvent& event) {
 			}
 			break;
 		}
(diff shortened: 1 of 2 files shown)


mercury233:
Do you think this was fixed in https://github.com/Fluorohydride/ygopro/commit/2a0cf61cfe41c2ff3c6a5fc2fd4f89b9731576b1 ?

zhykzhykzhyk:
Seems fixed

zhykzhykzhyk:
ddd

PR #1990, opened 2017-06-09
Create duelclient.cpp

Yumekaki:


edo9300:
Hmm, this could be interesting, aniways, why making 3 different pulls?

PR #1996, opened 2017-06-10
phase bottons

Yumekaki:


edo9300:
@Yumekaki the mr chech was changed, now you have to check for rule >=4 no longer 3

mercury233:
image
image
image
image

Would this be better?

PR #1998, opened 2017-06-12
update phase buttons

mercury233:
image
image
image
image


edo9300:
Where is the shuffle button?

_Sent from my Nubia NX529J using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

mercury233:
under the 3 chain buttons

Yumekaki:
distinguishing mr3 and mr4 is ideal

mercury233:
the DP button had become status indicator so it should be kept the same in MR3.

PR #2027, opened 2017-08-07
add nearest-neighbor anti-aliasing scaling for card pictures

mercury233:
`imageScaleNNAA` function taken from @Warr1024 in https://github.com/minetest/minetest

Scale the pictures with nearest-neighbor anti-aliasing algorithm, make it can read pics of any size perfectly. And no longer use thumbnail folder.

before vs after
image


Zayelion:
Looks ok to my newb eyes. Could use more comments.

PR #2032, opened 2017-08-22
Added Master Rule info in lan window

edo9300:


mercury233:
I prefer not to treat Master Rule specially, just show it as "Custom".

PR #2048, opened 2017-10-23
update mark for OCG/TCG cards

mercury233:
draw mark on cards which don't match the default OCG/TCG setting
image

if OCG/TCG search is selected, draw green mark on cards that match the default OCG/TCG setting
image


mercury233:
image
https://github.com/Fluorohydride/ygopro/pull/1858

PR #2051, opened 2017-10-24
Changed master rule options handling

edo9300:
With this change, i split all the different rules present in the various master rules in 5 "parameters", obsolete ignition, draw during the 1st turn, only 1 field spell on the field at the same time, pendulm zones and extra monster zones. Now the game will see the master rules as "Premade Packs" of those rules, allowing the player to choose freely among these 5 main options, still keeping the default ones.
master rule 4 parameters
image
some custom parameters
image
when joining a room with a "custom master rule" this window on the right will show up
image
otherwise it will display the master rule if different from the latest one.

ocgcore https://github.com/Fluorohydride/ygopro-core/pull/125


edo9300:
these are the strings added, i tried to translate them in chinese _but idk how good the translation is_
> !system 1264 Custom Rule  --自定义规则
> !system 1265 Obsolete Ignition --怪兽起动效果优先权
> !system 1266 1st Turn Draw --先攻抽卡
> !system 1267 1 Field --场上只能存在1张表侧表示的场地魔法卡
> !system 1268 Pendulum Zones --灵摆区域
> !system 1269 Extra Monster Zones --额外怪兽区域
> !system 1626 Custom --DIY

gframe/duelclient.cpp
```diff
@@ -334,8 +337,42 @@ void DuelClient::HandleSTOCPacketLan(char* data, unsigned int len) {
...
+		}
+		else if(pkt->info.duel_flag == MASTER_RULE_2) {
+			rule = 2;
+			mainGame->dInfo.duel_rule = 2;
+		}
+		else if(pkt->info.duel_flag == MASTER_RULE_3) {
+			rule = 3;
+			mainGame->dInfo.duel_rule = 3;
+		}
+		else if(pkt->info.duel_flag == MASTER_RULE_4) {
```
nekrozar: Official is using **New Master Rule**.
http://www.yugioh-card.com/japan/howto/new_master_rule/

DailyShana:
We only consider offical rules here.

youruriel:
Good idea, that will benefit Players who dont want to Play with Pendulum and Link, you have the Option to.

gframe/duelclient.cpp
Zayelion: None of the Konami rule sets use a consistent naming convention so refering to them by the English colloquialism is fine. There have been 4 rule sets.

Zayelion:
@DailyShana **They are official rules.** From a set time to a set time to recreate and simulate accurately tournament and game play settings such as the generation banlist.

Also "We only consider offical rules here." is nonsense because tag dueling is part of the system.

mercury233:
If we add this, `No Synchro` `No Xyz` `No Link` must be added too.
And I'm wondering if there is any tcg ruling that need core changes.

But personally I'll deal with this in future, not now.

DailyShana:
There are no such rules eg. that use obsolete ignition but do not draw in the first turn.
And it is tedious that add options for each newly added rule.
They are obsolete rules, even official do not maintain them, it is enough to surround the incompatible ones within `if(duel_rule...)` and let it be.

edo9300:
@DailyShana well if you put in that way, the old master rules were abandoned by konami, so everything should be removed and only the latest master rule should remain, so you should remove the obsolete ignition, the single field, the previous field and all those things too. For the "tedious" part, I don't see its problem, of a new rule gets added, it would still be needed to update the client/core, so it shouldn't be so tedious to add a parameter if a rule gets added, once you're already modifying the core/client.

cpcolella:
this is a great idea. I would also love the ability to select the ability to use xyz, synchro, link, etc as mercury233 says. There's no reason to not allow users to select their own custom rule sets.

PR #2053, opened 2017-11-04
Showing hand for tag partner

edo9300:
In a tag duel, you should be able to see the hand of your partner


DailyShana:
our tag duel follows the tf series, which do not allow to see the partner's hand.
npjh00142_00001

OhnkytaBlabdey:
about unshuffle mode , there's someone only expect unshuffle at the begin of duel, and shuffle during the duel.

mercury233:
so we should add all those rules to https://github.com/Fluorohydride/ygopro/pull/2051

Zayelion:
> our tag duel follows the tf series, which do not allow to see the partner's hand.

Says who? First was said IRL rules, then Tag Force rules, I am very sure TF is only an inspiration not a mandate. Konami issued tag rules contradic TF rules. Official OCG and TCG rules should have precedent.

StormWulf:
The official source of OCG Team Dueling is no longer available, but according to wikia http://yugioh.wikia.com/wiki/Team_Dueling_Rules#OCG_rules

> You are allowed to check and discuss your partner's hand and Set cards. 

> If an effect returns a card from the field, Graveyard, etc. to the hand or Deck, the card is returned to its original owner's hand/Deck. 

The latest official rules should have higher priority, even if they're not updated for MR4, but Tag Force is also obsolete.

PR #2071, opened 2017-12-14
Sound

mercury233:
Based on https://github.com/Fluorohydride/ygopro/pull/1662 .

Supported sounds are listed in sound/files.txt .

The sound manager now can only use irrklang, but can be updated to use other sound libraries, and irrklang is not required to build YGOPro.

To enable irrklang, let `USE_IRRKLANG` = true in premake.

Non-windows version is said to work (on mac), but not tested by me.


purerosefallen:
not supporting irrklang pro on this branch? i remember that irrklang pro have been used in your test branch and my kpro

PR #2077, opened 2018-01-06
update LoadExpansionDB and so on

nekrozar:
YGOPRO can check database, picture and so on in expansions subdirectory (Windows only).


nekrozar:
@purerosefallen I'm rewriting this code to support linux...
@mercury233 Shoud we use _zlib library_?

nekrozar:
https://github.com/Fluorohydride/ygopro/pull/2152 moved.

PR #2081, opened 2018-02-03
update lflist.conf

nekrozar:
TCG's Next Forbidden & Limited card lists
http://www.yugioh-card.com/en/limited/index.html


purerosefallen:
the banlist is not correct. there are a few mistakes.
for example, 27279764 should be limited

nekrozar:
https://github.com/Fluorohydride/ygopro/pull/2081/commits/8a569350adeeb262a4455dab5bb549befa4c721d fixed

PR #2088, opened 2018-02-15
Update to add Online mode similar to percy

John-cos:


John-cos:
Why was this closed?

mercury233:
It don't make any sense. Please finish your code before make pull request.

John-cos:
What it does is change the lan mode window so i basically functions like percys does. I modified it to get more data like name of duelists,type of duel etc etc.

mercury233:
Sorry, it is **code** which should be in pull request, not idea.

purerosefallen:
Why should we be similar to Percy? It really makes 0 sense.

John-cos:
There is no reason it shouldn't be a standard feature.

PR #2090, opened 2018-02-20
update directly open ydk and yrp file

mercury233:
make ygopro can directly open ydk and yrp file
e.g. 
`"ygopro.exe" "F:\deck\1.ydk"`

### DIFF
--- a/gframe/gframe.cpp
+++ b/gframe/gframe.cpp
@@ -157,20 +157,22 @@ int main(int argc, char* argv[]) {
 			if(open_file)
 				ClickButton(ygo::mainGame->btnLoadSinglePlay);
 			break;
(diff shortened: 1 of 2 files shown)


gframe/gframe.cpp
```diff
@@ -156,6 +157,21 @@ int main(int argc, char* argv[]) {
 			if(open_file)
 				ClickButton(ygo::mainGame->btnLoadSinglePlay);
 			break;
+		} else if(argc == 2) {
+			char* pstrext = argv[1] + strlen(argv[1]) - 4;
```
VanillaSalt: if strlen(argv[1]) < 4 access error may occur

PR #2104, opened 2018-04-17
fix select unselect

mercury233:
https://github.com/Fluorohydride/ygopro-core/pull/146
the only difference between `finishable` and `cancelable` is the former will show the finish button instead of cancel button.

### DIFF
--- a/gframe/duelclient.cpp
+++ b/gframe/duelclient.cpp
@@ -1397,8 +1397,9 @@ int DuelClient::ClientAnalyze(char * msg, unsigned int len) {
 	}
 	case MSG_SELECT_UNSELECT_CARD: {
 		/*int selecting_player = */BufferIO::ReadInt8(pbuf);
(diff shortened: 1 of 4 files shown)


gframe/duelclient.cpp
```diff
@@ -1397,8 +1397,9 @@ int DuelClient::ClientAnalyze(char * msg, unsigned int len) {
 	}
 	case MSG_SELECT_UNSELECT_CARD: {
 		/*int selecting_player = */BufferIO::ReadInt8(pbuf);
-		bool buttonok = !!BufferIO::ReadInt8(pbuf);
-		mainGame->dField.select_cancelable = BufferIO::ReadInt8(pbuf);
+		bool finishable = BufferIO::ReadInt8(pbuf);
+		bool cancelable = BufferIO::ReadInt8(pbuf);
```
purerosefallen: this will give out a warning for converting int into bool, why not `!!BufferIO::ReadInt8(pbuf);` ?

PR #2107, opened 2018-04-19
add disable chat

mercury233:
When the chat of opponent is ignored, the player himself will also be muted.


purerosefallen:
Why to mute the player themselves?

PR #2109, opened 2018-04-21
Updated descriptions to 64 bit values

edo9300:
Since lua 5.3 natively supports 64 bit values, in this way, the range of strings useable for the descriptions increase, and allow using descriptions with cards with 9+ digit ids without problems.
core ralated changes https://github.com/Fluorohydride/ygopro-core/pull/149


DailyShana:
unnecessary

PR #2111, opened 2018-05-06
add screenshot

purerosefallen:
press key `X` or `F12` to take a screenshot, which will save at `/screenshots`.


jimmyvordick:
I use f12 and force close 🙂🙏

PR #2118, opened 2018-06-25
add SrvPro Menu

nekrozar:


edo9300:
I think you should do a check and see if there's at least 1 server listed in the configs, otherwise hide the server tab

purerosefallen:
In addition, We need to add an option to let users add servers.

purerosefallen:
I had a quick test of it, but it crashed after enter a server.

purerosefallen:
user gone
and commit broken
and not maintainable

PR #2125, opened 2018-07-31
update loading expansion script

DailyShana:
do not need ocgcore to consider where to load expansion scripts and more flexible to extend for reading scripts in other directories or even in zips.


DailyShana:
In single mode, when reading puzzle scripts, using `WideCharToMultiByte` is more direct and simple, and it's nothing to do with MBDS programming although it use a little multi-byte character because of the behavior of `fopen` in Windows systems.

purerosefallen:
A better idea: `load_script("c????????")`

DailyShana:
for compatibility reasons, `load_script("./script/c%d.lua")` should not be changed.

PR #2126, opened 2018-07-31
add new style of select option

mercury233:
image
image

Long options are still using the old style of dialog.


purerosefallen:
I approve this one.

PR #2128, opened 2018-08-07
add skip action

purerosefallen:
Press Z to skip any animations to the current turn. This works as duelists or observers in duels, but not works in watching replay.


edo9300:
I don't think this feature is needed

purerosefallen:
@mercury233 then you would see a card teleported from one place to another, having no idea of what had happened or which effect your opponent used.

PR #2131, opened 2018-08-14
Allow select option quick to select among more than 5 values

edo9300:


edo9300:
If tehre are more than 5 options to choose from, and all those options' texts can enter into the buttons a scrollbar will be added to allow you to better choose among them 
image

mercury233:
I didn't test it but shouldn't we reset `scrOption` in `ShowSelectOption` ?

edo9300:
wait..., damn github desktop.. it didn't push all teh files

gframe/client_field.cpp
```diff
@@ -594,26 +594,34 @@ void ClientField::ShowSelectOption(int select_hint) {
...
-		if(dtxt.Width > 310) {
+	for(auto option : select_options) {
+		if(mainGame->guiFont->getDimension(dataManager.GetDesc(option)).Width > 310) {
 			quickmode = false;
 			break;
 		}
+	}
+	for(int i = 0; (i < count) && quickmode; i++) {
+		const wchar_t* option = dataManager.GetDesc(select_options[i]);
 		mainGame->btnOption[i]->setText(option);
```
VanillaSalt: out of bound if count > 5

purerosefallen:
does it work fine with resize?

edo9300:
@purerosefallen i remember you that that still hasn't been merged...

purerosefallen:
@edo9300 but did you change anything else on EDOPro?

Issue #2132, opened 2018-08-24
Issues about MSG_CONFIRM_CARDS


purerosefallen:
Now MSG_CONFIRM_CARDS just sends a single message to make the client show cards. But in panel mode (when confirming 2 or more cards in deck or extra deck) , the client does not do anything in the confirming dialog until player clicks OK. The appearance of this issue is, the timer stops at confirming opponent's extra deck when activating Reaper & Winter Cherries.
This is sometimes used in a FOUL PLAY, activating Ghost Reaper & Winter Cherries, then simply wasting time at the dialog confirming opponent's extra deck, then wait until the extra duel starts. This is somehow serious because it may have bad influences on tournaments.
A solution to this is, to make MSG_CONFIRM_CARDS work same with other MSGs, like MSG_SELECT_CARD, making it going into `WaitforResponse` after MSG_CONFIRM_CARDS and continue the duel only after the client sents CTOS_RESPONSE.
PRs are welcomed if you could make this change, thanks.

PR #2133, opened 2018-08-24
check timeout when waiting CTOS_TIME_CONFIRM

DailyShana:
for #2132
Not only `MSG_CONFIRM_CARDS`, all msg except `MSG_SELECT_*` have this issue as long as the client hang on when solving these msg intentionally.
The elapsed time do not count in the time limit of the player.


purerosefallen:
Does it mean, I will lose my duel when activating Reasoning with only 1 min left?

edo9300:
I think this will give advantage to people with slightly different clients, with like unlocked fps or skipped animations, because the messages will take less to execute

purerosefallen:
@edo9300 *So why don't you duel on shell*

DailyShana:
I think it can be done easier in actual server code, eg. svrpro, ygosharp, etc.

PR #2134, opened 2018-08-28
Add announce card cache

purerosefallen:
When you announce a card for the 2nd time or more this match, the announced card will be displayed at the first page.
a1
a3
a2


DailyShana:
#2137 implement this in a simple way

PR #2135, opened 2018-09-03
Added support for messages in the puzzle selection window

edo9300:
With this, by butting a specific header in the script of a puzzle, the client will be able to read a message from it 
image

the header should be put in this way
```
#header
things that will be shown
#endheader
lua script
```

this commit is required first https://github.com/Fluorohydride/ygopro-core/pull/185


DailyShana:
it breaks the nature of lua script.
It can be done by reading the comment of the script in the first lines.

PR #2138, opened 2018-09-05
add export deck from replay

purerosefallen:
qq 20180905102357
Added the function of exporting decks from a replay file.
~~Don't use this to cheat~~


mercury233:
I don't think any feature should be add to the replay system before we add the new replay system.

purerosefallen:
@mercury233 According to Shana's words, this is compatible with the new replay, because the new replay also includes both players' decks at the first few bytes. Only a few lines will be required to change after the new replay system is finished.

DailyShana:
The planning new replay can view both players' decks but do not need to record decks at the first few bytes.
The deck data is in the MSG_UPDATE_DATA.
The server use `RefreshDeck` to save deck data for replay and never send it to players.

PR #2139, opened 2018-09-19
add quick animation

mercury233:
make all waiting no longer than 0.2s


purerosefallen:
Why the string 1298 is used twice?

PR #2140, opened 2018-09-19
update setting working directory

DailyShana:
It disable the possiblility to run program in other working directory.

So set working directory only when opening files.


mercury233:
but opening file from double clicking file in explorer require this

edo9300:
> but opening file from double clicking file in explorer require this

@mercury233 i don't see what the issue could be, i tested and the client works just fine even without setting the workign path

mercury233:
@edo9300 Windows set the working dir to the dir of the file being double clicked, instead of the dir of the exe used to open the file.

PR #2141, opened 2018-09-19
 update parse param

mercury233:


gframe/gframe.cpp
```diff
@@ -59,112 +56,60 @@ int main(int argc, char* argv[]) {
...
-		}
-		++ptr;
-	}
-	if(in_TEXT) {
-		argv[argc][j] = '\0';
-		++argc;
+	int wargc;
+	LPWSTR commandLine = GetCommandLineW();
+	wargv = CommandLineToArgvW(commandLine, &wargc);
+#else
```
edo9300: You should free the returned buffer after you finish working with it

gframe/gframe.cpp
```diff
@@ -59,112 +56,60 @@ int main(int argc, char* argv[]) {
...
+	for(int i = 1; i < argc; ++i) {
+		BufferIO::DecodeUTF8(argv[i], wargv[i]);
```
DailyShana: writing directly to argv[i] will cause buffer overflow

PR #2142, opened 2018-09-27
remove cancel button hint

mercury233:
It will do nothing but confusing the users and disable useful feature.


purerosefallen:
~~The hint button covered my wallpaper's sexual bodypart so I turned it off~~

PR #2144, opened 2018-09-29
add surrender confirm

mercury233:
image


purerosefallen:
What if clicking Surrender when selecting a card or performing other actions? Will the original dialog disappear?
In addition, what if you are asked to select a card when confirming a surrender?

mercury233:
@purerosefallen It is using a separate dialog(window), so it won't affect any other actions.

PR #2145, opened 2018-09-29
update "activate" and "become target" effect for cards in grave or removed

mercury233:
target:
gif3
gif4

activate:
gif5

in quick mode:
[gif](https://user-images.githubusercontent.com/13391795/46249256-b1907280-c458-11e8-9fe6-548ad671edeb.gif)

require https://github.com/Fluorohydride/ygopro-core/pull/188

PR #2146, opened 2018-09-30
add tip for cards in select / display list

mercury233:
image
image

_make https://github.com/Fluorohydride/ygopro/pull/2145 works better_


purerosefallen:
Can it work without #2145?

PR #2147, opened 2018-10-01
build lua as c++

DailyShana:
another solution of Fluorohydride/ygopro-core#186


purerosefallen:
Should we cut out `os` and `io` libs in OUR lua source? This will prevent harmful lua scripts (especially custom cards) destroying the server.

PR #2149, opened 2018-10-23
update pos of ShowCardInfoInList, fix the tip blink

mercury233:
image
image
image


purerosefallen:
I don't think there are super long cards names, at least in OCG/TCG.

mercury233:
image

purerosefallen:
image
why not expanding the tip size here?

mercury233:
image
it cant go beyond the window

edo9300:
Wouldn't putting the tip object outside the card select solve that?

mercury233:
@edo9300 it will be a bit ugly imo... 
image
image

PR #2151, opened 2018-10-27
place cards which are activating or target in first when selecting cards

mercury233:
image


purerosefallen:
So what is the meaning of the previous works making the cards not moving in grave?

edo9300:
In this way it's cleaner for the user without going against the rules moving the cards physically

PR #2158, opened 2018-11-17
Search cards with RegEx

purerosefallen:
Now you can use any RegEx in card search. Cards matching given RegEx will be at the first.


edo9300:
@purerosefallen please, remove that doubled definition that's basically the same stuff (it's weird that the compiler didn't detect it as ambiguous) also take the strings by reference as const values

purerosefallen:
where's the double definition?

DailyShana:
I dont think there is any necessity so support regex. It's too professional for general users.

cpcolella:
on the flip side, not including RegEx is too basic for advanced users.

NewScript:
E pelo lado que estão esquecendo, usar regex vai adicionar alguma experiência para o usuário?

PR #2160, opened 2018-11-20
use windows api instead of c runtime library in myfilesystem.h

mercury233:
There is a known bug in UCRT, and it is much easier to switch to another way than fixing this bug in UCRT way.

https://developercommunity.visualstudio.com/content/problem/19317/windows-c-function-wstat-does-not-work-correctly-w.html


DailyShana:
The headers for those functions can be removed

PR #2164, opened 2018-12-02
Multi language

purerosefallen:
Supports multi languages which can be changed in settings.
The directory structure is the same with https://github.com/moecube/ygopro-database. Just toss in the whole `locales` directory and it will work.
Currently, `pics`, `strings.conf`, `cards.cdb`, and `bot.conf` are supported for multi language.


purerosefallen:
In addition, at the 1st run of ygopro, it will detect the system language for the default language.

PR #2165, opened 2018-12-07
use one ScriptReaderEx

mercury233:
It will slightly decrease the performance on Windows because `DecodeUTF8` is always called


purerosefallen:
I don't think it important, unless on server. In addition, now servers are almost all on Linux

edo9300:
@purerosefallen so better keep the code all messed up only because the server are on Linux, interesting point of view

PR #2178, opened 2019-03-07
Reverse Log Order

InvisibleMoon:
Always show the newest battle state in front of the list box, which might be helpful for identifying what happened during the game immediately.
Linked with Issue 2175, and it will be closed later.


Smile-DK:
Better add an option to give the choice to players.

edo9300:
A log in reverse order would be unpractical to use as you won't properly follow the flow of the events, as they'll be in reverse order

InvisibleMoon:
> Better add an option to give the choice to players.

You are right. I didn't do that because it would involve states between transition and increase complexity, but player always the first.

But I'd to explain that reverse log wouldn't be unpractical to use , at lease on a broad view. 
Player wants to check cards info at the time it happens, not with history. Because people can remember events they already understood, but unknown one's harder. 
In the scene, "event flow" is what should be offered, not the "log". 
If someone wants to check actions before, by his accidental need, he could always scroll down at the corrent location to determine what's on.

mercury233:
I think https://github.com/Fluorohydride/ygopro/pull/2194 will be a better solution

PR #2181, opened 2019-03-27
update MSG_START

DailyShana:
for preparing to test #2179


edo9300:
Was there really a need for this?

PR #2182, opened 2019-04-12
add deck code

purerosefallen:
QQ图片20190413011756
QQ图片20190413011750

The deck code is stored in base64 with the format of MSG_UPDATE_DECK, in convenience of sharing decks in the community.


DailyShana:
There is already an implement of `deck://` protocol: 
https://github.com/Fluorohydride/ygopro/blob/ygopro2/netc/deck_data.cpp#L122
https://github.com/Fluorohydride/ygopro/blob/ygopro2/netc/deck_data.cpp#L192
which can be used copy/paste decks directly.

mercury233:
If the user must copy the data to clipboard, there are no reason to use unreadable code instead of deck text.

PR #2183, opened 2019-04-14
auto respond MSG_SELECT_CHAIN when only 1 mandatory effect avail

mercury233:


salix5:
我支持保留現在的設定，主要的理由是：
1. 
"在某個階段中符合發動條件的只有1個必發效果"表示手牌等非公開區域沒有可以發動的選發效果
這是只有玩家A自己知道的情報，不應該讓玩家B過觀察系統的處理動作而察覺

具體的例子：
A的墓地有被破壞的炎王神獸，手牌沒有增殖的G，進入B的準備階段
目前：B的準備階段中，A選擇墓地的炎王神獸發動效果
修改後：B的準備階段中，A的炎王神獸自動發動效果

A的墓地有被破壞的炎王神獸，手牌有增殖的G，進入B的準備階段
目前：B的準備階段中，A選擇墓地的炎王神獸發動效果
修改後：B的準備階段中，A選擇墓地的炎王神獸發動效果

也就是說在修改後
B可以通過觀察"A的炎王神獸是不是自動發動效果"得知"A的手牌有沒有增殖的G等自由時機的手坑"

2. 
階段中的必發類誘發效果和一般時機的必發類誘發效果有本質上的不同
一般時機的必發效果：
玩家必須發動這個效果才能進行下一個動作，行動順序有嚴格限制
因此一定程度的自動化並不影響整體的流程

階段中的必發效果：
玩家必須發動這個效果才可以結束這個階段
但發動的順序並沒特別的限制
玩家可以發動任意數量的其他選發效果，最後發動必發效果

以這點來說，必發效果應該和選發效果採用一樣的處理方式
即使只有1個符合條件的必發效果
玩家也應該比照只有1個符合條件的選發效果的情形，照常選擇那張卡發動效果

I think the current setting is more appropriate
The main reasons are:
First, the fact that "There is only 1 mandatory effect available in this phase." includes information from non-public area, and the opponent shouldn't know this from the game system.

Example:
A's GY: a destroyed Fire King High Avatar Garunix
A's hand: no Maxx "C"
in B's Standby Phase
Now: A chooses Garunix and activate its effect
Modified: Garunix's effect is activated automatically

A's GY: a destroyed Fire King High Avatar Garunix
A's hand: 1 Maxx "C"
in B's Standby Phase
Now: A chooses Garunix and activate its effect
Modified: A chooses Garunix and activate its effect

That is, B can know that A does not have a Maxx "C" if Garunix's effect is activated automatically.

2. Mandatory effects in ordinary timing and mandatory effects in phase are different
Mandatory effects in ordinary timing:
The player must activate the effect in order to do the next move, so there is a specific order.
Some sort of automation is fine since the player cannot do anything else.

Mandatory effects in phase:
The player must activate the effect before the phase ends, but the order is not specific.
The player can activate any number of optional effects before the mandatory effects.

Therefore, I believe that mandatory effects in phase should have the same process as optional effects.
Even if there is only 1 mandatory effect available, the player should choose that card and activated its effect, just like what one does when there is only 1 optional effect available.

PR #2185, opened 2019-04-15
ydk:// and copy/paste decks

DailyShana:
from https://github.com/Fluorohydride/ygopro/tree/ygopro2 branch
string 1368 is `The deck has been copied to clipboard`


purerosefallen:
Could you change the deck code format to MSG_UPDATE_DECK like https://github.com/Fluorohydride/ygopro/pull/2182 ?
In KoishiPro2 iOS, players have saved many deck codes for a week. In addition, There are plans in MyCard that deck codes in the format of the above PR would be posted.
Creating another format of deck code would make the players complain why their deck codes are incompatible.

DailyShana:
just keep your format in your branch for compatibility, which is only valid for your users, not all users
if you had discussed this feature request with us, there would be no such problem.

purerosefallen:
*Could you implant this version of deck code into ygopro2?*

purerosefallen:
In addition, the maximum card code should be 0xfffffff, rather than 0x7ffffff. It may break the compatibility other cards in expansions.

*Did you really fully reviewed the ygopro2 branch in fh?*

edo9300:
If you look carefully, the "extra space" is used to store the card count and there's the 0x2000000 flag for cards in the side deck

purerosefallen:
@edo9300 I saw the use of the extra space. But it occupied 5 bits. Only 4 bits are available however.

DailyShana:
`0x7ffffff=134217727` is 9 digits but official cards has only 8 digits so there is no restriction.
and in addition it just happens that `DataManager::GetDesc()` use **signed** integer as parameter, so the max available card id is just `0x7ffffff=134217727`

purerosefallen:
Pre-release cards may go over this value.
In addition, current max card code is 0xfffffff in the current ygopro version. It is not very acceptable to shrink the maximum card code possible, just for compating the card code.

edo9300:
@DailyShana i was thinking, wouldn't it be better to jsut store the extra, main, and side count, and just put all the cards in order without caring about duplicated optimization? after all the gain from such "compression" is pretty minimal and can be easily ignored, and this will allow this "encoding" to store a whole int32

gframe/deck_con.cpp
```diff
@@ -675,6 +675,30 @@ bool DeckBuilder::OnEvent(const irr::SEvent& event) {
...
 	}
 	case irr::EET_KEY_INPUT_EVENT: {
+		if(event.KeyInput.PressedDown && event.KeyInput.Control) {
+			switch(event.KeyInput.Key) {
+			case irr::KEY_KEY_C: {
+				if(!mainGame->HasFocus(EGUIET_EDIT_BOX)) {
+					std::wstring deck_string = L"ydk://";
+					deck_string += deckManager.SaveToString(deckManager.current_deck);
+					mainGame->device->getOSOperator()->copyToClipboard(deck_string.c_str());
+					mainGame->stACMessage->setText(dataManager.GetSysString(1359));
```
DailyShana, line 685:
```suggestion
					mainGame->stACMessage->setText(dataManager.GetSysString(1368));
```

PR #2198, opened 2019-05-18
remove unnecessary typing enter key when declaring card name

DailyShana:


mercury233:
it is better to use `gameConf.auto_search_limit`

DailyShana:
auto searching in declaring name have exist for a long time, it's better to keep it without condition

PR #2199, opened 2019-05-26
update MSG_ANNOUNCE_CARD_FILTER

mercury233:
The count sent by the core is in unsigned int.


purerosefallen:
In fact many msgs have the same problem..

PR #2217, opened 2019-06-19
do not stop server at duel end process

### LINKED ISSUE #2067: Connection has been lost before winning / losing
Hello, it suddenly stopped working :P And I do not know what caused the error.

When a player has to win / lose a duel (lose LP), players are disconnected without a final result. Maybe someone knows the solution to this? Unfortunately, I have no information what causes this problem.

Thank you for your answer

DailyShana:
server will stop when host player leaves the game
fix #2067

### DIFF
--- a/gframe/network.h
+++ b/gframe/network.h
@@ -114,7 +114,7 @@ struct DuelPlayer {
 
 class DuelMode {
 public:
-	DuelMode(): host_player(0), pduel(0) {}
+	DuelMode(): host_player(0), pduel(0), duel_stage(0) {}
 	virtual ~DuelMode() {}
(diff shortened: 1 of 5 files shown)


DailyShana:
when leaving game after duel end, sending `MSG_WIN` with reason lost connection is still triggered

DailyShana:
when all players leave game after duel end, server will stop automatically
even [this line](https://github.com/Fluorohydride/ygopro/pull/2217/files#diff-de360d2161164ea2b9e41125930dab07L139) is deleted, 
but then server will not stop if all player leave game before duel start,
so for simplicity, just keep it as before

PR #2218, opened 2019-06-21
Clean up broadcasting code

edo9300:
Now when hosting a game with the client, the server will no longer use 3 ports to send the duel info when breadcasting, but will do everything on the port the duel is currently hosted on, also some redoundancies have been removed.

### DIFF
--- a/gframe/duelclient.cpp
+++ b/gframe/duelclient.cpp
@@ -3809,45 +3809,24 @@ void DuelClient::BeginRefreshHost() {
 	if(!host)
 		return;
 	SOCKET reply = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
-	sockaddr_in reply_addr;
-	memset(&reply_addr, 0, sizeof(reply_addr));
-	reply_addr.sin_family = AF_INET;
-	reply_addr.sin_port = htons(7921);
-	reply_addr.sin_addr.s_addr = 0;
-	if(bind(reply, (sockaddr*)&reply_addr, sizeof(reply_addr)) == SOCKET_ERROR) {
-		closesocket(reply);
-		return;
-	}
+	BOOL opt = TRUE;
+	setsockopt(reply, SOL_SOCKET, SO_BROADCAST, (const char*)&opt, sizeof(BOOL));
 	timeval timeout = {3, 0};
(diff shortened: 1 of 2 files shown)


DailyShana:
The port that the duel is hosted on is not necessarily `7911`

edo9300:
In the client it's currently hardcoded to that

DailyShana:
see
https://github.com/Fluorohydride/ygopro/blob/master/gframe/game.h#L18
https://github.com/Fluorohydride/ygopro/blob/master/system.conf#L12

edo9300:
IK, that, but for a client it's still unknown which port the other client is hosting on, now I could have made it use the port in the ip address when refreshing, but probably that wouldn't have been ok for your "standards"

DailyShana:
because you says
>will do everything on the port the duel is currently hosted on

in fact, I'm not against hardcoding 7911

purerosefallen:
But it would be horrible if under NAT...
I have to use iptables to transfer the traffic to another port.

edo9300:
@purerosefallen do you realize tbat currently this code isn't using only 1 hardcoded port, but 2?

gframe/duelclient.cpp
```diff
@@ -3809,45 +3809,24 @@ void DuelClient::BeginRefreshHost() {
...
-			break;
-		BOOL opt = TRUE;
-		setsockopt(sSend, SOL_SOCKET, SO_BROADCAST, (const char*)&opt, sizeof(BOOL));
-		if(bind(sSend, (sockaddr*)&local, sizeof(sockaddr)) == SOCKET_ERROR) {
-			closesocket(sSend);
-			break;
-		}
-		sendto(sSend, (const char*)&hReq, sizeof(HostRequest), 0, (sockaddr*)&sockTo, sizeof(sockaddr));
-		closesocket(sSend);
+		sockTo.sin_addr.s_addr = local_addr;
```
DailyShana: it only connect to the address of the native device, cannot get the remote host in the LAN

edo9300: @DailyShana what do you mean? In the various tests i made, this code behaved exactly like the previous one with no differences at all, what's the non working setup?

DailyShana: in the LAN, one pc set up a host, the other pc cannot get it by refreshing host

PR #2222, opened 2019-07-11
fix showing alias

mercury233:
for custom cards whose codes may be near


purerosefallen:
They all know the difference `10` of the aliases...
In addition, how about the ocgcore side?

edo9300:
> In addition, how about the ocgcore side?

This doesn't have to be handled in the core, from what i understand it only makes the double name not show if the original is the same as the aliased one

PR #2226, opened 2019-07-16
core update!!!

gianni159:
can the guys that do the core update for ygopro please update the core so the scripters can script rousing hydradrive and the other cards that need an update like judgment arrows and some monsters please we all would be very gratefull if you update the core of ygopro


mercury233:
only cards which actually exist can be considered

PR #2241, opened 2019-11-21
fix problems with building on macOS

Liu-Junyan:
1. Change "GL" prefix in include path to "OpenGL" for macOS.
2. Include Cocoa and IOKit in external libraries list as they are required by Irrlicht engine on macos.
Built successfully with cmake on macOS 10.15.1. However, premake scripts have not been modified, 
so corresponding tweaks need to be made in order to allow build with premake.

### DIFF
--- a/gframe/config.h
+++ b/gframe/config.h
@@ -56,8 +56,6 @@ inline int myswprintf(wchar_t(&buf)[N], const wchar_t* fmt, TR... args) {
 }
 
 #include <irrlicht.h>
-#include <GL/gl.h>
-#include <GL/glu.h>
 #include "CGUITTFont.h"
 #include "CGUIImageButton.h"
 #include <iostream>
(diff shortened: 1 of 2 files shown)


kevinlul:
These have been resolved in many downstream forks but non-Windows platforms don't seem to be a priority for upstream.

gframe/config.h
```diff
@@ -56,8 +56,6 @@ inline int myswprintf(wchar_t(&buf)[N], const wchar_t* fmt, TR... args) {
 }
 
 #include <irrlicht.h>
-#include <GL/gl.h>
-#include <GL/glu.h>
```
DailyShana: just change in place, do not move to other place and do not add extra empty lines

DailyShana:
cmake have official `CMakeFindFrameworks` module
see https://github.com/Kitware/CMake/blob/master/Modules/CMakeFindFrameworks.cmake
I feel using this is more appropriate but I have no condition to test.

DailyShana:
merged in 293618b

PR #2273, opened 2020-03-21
Ygopro2

Nexus4321:
I'm kinda new to this and I'm not so sure if I'm doing this right or not, but is there any way to switch out the old files for a new one it's for card called Clear World? The reason why it's because it's not the right Card Art it's the Anime one I believe and not the actual TCG/OCG. A staff member from dueling nexus told me to come over here for a request so here's a link to the actual image of the card that needs to be switched into it, thank you :D https://images.app.goo.gl/8nguvpaJn52UYRB2A


mercury233:
you didn't download the card pictures from here, so here isn't the right place to report.

Nexus4321:
If this isn't the right place to report what will I have to do? Also if anyone is talking to me then say Nexus so I would at least know, sorry I'm just new to this stuff :/

SoaringSky:
Hi, this is Soaring__Sky from Dueling Nexus. I don't know who referred you to this place, but like Mercury said, the card pictures don't come from here. This is a client-related issue, so if anything, you should ask me about it, not bother the people on Fluoro about it. I can tell you we're working on a way to change how our card art is displayed, but no guarantees yet.

PR #2279, opened 2020-04-09
Use environment variables instead of working directory

LordYuuma:
Currently, all YGOPRO builds load data from the working directory, irrespective of the OS they run in.  This patch adds environment variable support mostly for Linux builds, so that the binary can be run anywhere (especially `$prefix/bin`) while still loading files from well-known locations.

`--environment-paths` enables
1. loading card databases, lflists, etc. from `YGOPRO_DATA_PATH`,
2. loading images from `YGOPRO_IMAGE_PATH`, and
3. loading scripts from `YGOPRO_SCRIPT_PATH`.

`--xdg-environment` enables 
1. loading user configuration from `$XDG_CONFIG_HOME/ygopro (~/.config/ygopro)` or `$prefix/etc/ygopro`, and
2. loading decks, images, sounds, replays and singles from `$XDG_CONFIG_HOME/ygopro (~/.local/share/ygopro)` or `$prefix/share/ygopro`.

A variant of this patch was sent to mycard as https://github.com/mycard/ygopro/pull/9.

### DIFF
--- a/gframe/config.h
+++ b/gframe/config.h
@@ -54,7 +54,6 @@ template<size_t N, typename... TR>
 inline int myswprintf(wchar_t(&buf)[N], const wchar_t* fmt, TR... args) {
 	return swprintf(buf, N, fmt, args...);
 }
-
 #include <irrlicht.h>
 #ifdef __APPLE__
 #include <OpenGL/gl.h>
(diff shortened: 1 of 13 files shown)


purerosefallen:
not maintainable

mercury233:
We do need to add some features like this in the future.

PR #2280, opened 2020-04-19
support new expansion file extension

mercury233:
~~.yxp (YGOPro eXPansion)~~
.ypk (YGOPro PacK)

make it less confusing for user to install expansion

### DIFF
--- a/gframe/game.cpp
+++ b/gframe/game.cpp
@@ -935,15 +935,15 @@ void Game::LoadExpansions() {
 			myswprintf(fpath, L"./expansions/%ls", name);
 			dataManager.LoadDB(fpath);
 		}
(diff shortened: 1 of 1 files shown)


kevinlul:
Does obfuscating the fact that an expansion pack is a ZIP archive really make things _less_ confusing?

mercury233:
You can directly tell the user to put .ypk file in the expansions folder. If it is .zip file, the user may wonder should he extract it, and some user may put other .zip which don't have correct structure in the expansions folder.
We can also associate .ypk with ygopro so the user can directly double click on it to install.

Renamed .zip extension is wildly used, such as .docx, .apk, .ipa, .epub, .jar, .crx

PR #2283, opened 2020-04-21
auto trim text after pasted

mercury233:


mercury233:
{{MH{14GF3FJV0F6OEID {R
_we do still need this feature_

purerosefallen:
How about this one

mercury233:
_I forget what was the problem_

purerosefallen:
> _I forget what was the problem_

When pasting
```
s1.ygo233.com

```

the textbox would add a newline, and make it fail to resolve.

mercury233:
_I mean the reason I turned this to draft_

purerosefallen:
which of the textboxes in ygopro are multiline, and which not?

PR #2289, opened 2020-04-28
add support for running external executable on startup

mercury233:
It can be useful for something like update checker


DailyShana:
also useful for something like trojan?

mercury233:
If the malware can modify system.conf, it can modify the main ygopro exe too, so this feature won't increase any risk.

edo9300:
Rather than making the client spawn processes, the safer approach would be that the other process spawned the client.

mercury233:
@edo9300 It won't be safer, and may require all old users to change the shortcut.

salix5:
I think that there is a security concern too.
`system.conf` is a text file which can be modified easily, and some users might get the config file from some unknown sources.

Asking the users to run an update checker by themselves might be sort of awkward, but it is much safer.

PR #2292, opened 2020-05-06
add chain hint for cards on field

mercury233:
image

If one player have two cards of same name and location, one of them activate effect, the opponent won't know which card is activating if there is only 1 chain.

### DIFF
--- a/gframe/event_handler.cpp
+++ b/gframe/event_handler.cpp
@@ -1590,6 +1590,18 @@ bool ClientField::OnEvent(const irr::SEvent& event) {
 									str.append(formatBuffer);
 								}
 							}
+							for(size_t i = 0; i < chains.size(); ++i) {
+								wchar_t formatBuffer[2048];
+								auto chit = chains[i];
+								if(mcard == chit.chain_card) {
+									myswprintf(formatBuffer, dataManager.GetSysString(216), i + 1);
+									str.append(L"\n").append(formatBuffer);
+								}
+								if(chit.target.find(mcard) != chit.target.end()) {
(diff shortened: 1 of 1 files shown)


DailyShana:
This is redundant, as there are textures `tChain` and `tChainTarget` showed.
The problem is that `tChain` is showed for chain 1 only when chain 2 is activated, so just remove this setting, showing `tChain` texture for chain 1 as soon as it is activated is enough

mercury233:
> showing tChain texture for chain 1 as soon as it is activated

It will be ugly

DailyShana:
> > showing tChain texture for chain 1 as soon as it is activated
> 
> It will be ugly

???
tChain is just for this usage, showing a card is activated in which chain.

mercury233:
I don't think the users will like the chain mark always showing

DailyShana:
> I don't think the users will like the chain mark always showing

They want to know which card is activated, but don't like showing chain mark? Moving mouse to find out which card is activating?

mercury233:
The dialog will show the name of card activated

DailyShana:
Still need to move mouse to find which card is activating when two cards of same name on field

DailyShana:
summary
all information is in chain mark and chain target mark, showing chain hint for cards on field is unnecessary.
the only missing information is don't know which card is activating when only chain 1 is activated, so the solution is showing chain mark for chain 1

mercury233:
it is very ugly if always showing chain 1 make, and it is weird if only show chain 1 mark if there are card of same name

DailyShana:
how does it ugly? this is just unrelated subjective point.
Showing which card is chaining is necessary, then just showing chain mark.

PR #2293, opened 2020-05-06
ignore STOC_GAME_MSG outside duel

mercury233:
Sometimes the client still receive `STOC_GAME_MSG` when changing side, and processing that message may crash the client because the field is cleared.


DailyShana:
sometimes is what time? the server side should guarantee the correctness of messages

mercury233:
_if the order is guaranteed, the client can safely ignore that msg_

DailyShana:
what order?
If the client receives `STOC_GAME_MSG` when changing side, there must be a bug in the server

purerosefallen:
There is 1 case. Sometimes `MSG_WIN` would be received while changing side when opponent quits.

PR #2296, opened 2020-05-15
update time limit handling

mercury233:
The time limit timer shouldn't rely on `CTOS_TIME_CONFIRM`, as the client can delay the respond to get more thinking time.


mercury233:
It is used on my server for years

PR #2187, opened 2019-04-21
Deck category and deck manage

mercury233:
image

PR #2297, opened 2020-05-18
Folder and Sub Folder in Deck List

jimmyvordick:
I hope, next, Edopro can read the folders and sub folders on the decklist.  I understand that in the future there will be many types of decks that we must save ... such as the Master Rules Deck, Speed ​​Deck, RUSH deck, anime deck, and others.  I like Replay which can now read folders and sub folders.  I hope the deck list is the same.  can read folders and subfolders.  thank you


mercury233:
https://github.com/Fluorohydride/ygopro/pull/2187

PR #2299, opened 2020-05-20
distinguish chain 1 if the activated card has same name cards on field

mercury233:


DailyShana:
can't agree this change, there is no problem to solve

PR #2305, opened 2020-06-05
update font glyph page

mercury233:
The `CGUITTFont` store the glyph of font in pages, a 256x256 page have many sprites of char in font_size*font_size. But the size of sprite may be larger than font_size(eg. `|`), so the sprite may overwrite other sprites or be overwritten.
So we should enlarge the size of sprite.


edo9300:
What's a before/after with this change?

mercury233:
image
With some font, here is overwritten by other char.

PR #2308, opened 2020-07-04
update conti_act operation mark

mercury233:
GIF

The conti_act operation mark may be barely noticeable on some background images, so draw the cards behind it.

### DIFF
--- a/gframe/drawing.cpp
+++ b/gframe/drawing.cpp
@@ -452,9 +452,21 @@ void Game::DrawMisc() {
 		driver->drawVertexPrimitiveList(matManager.vActivate, 4, matManager.iRectangle, 2);
 	}
 	if(dField.conti_act) {
-		im.setTranslation(vector3df((matManager.vFieldContiAct[0].X + matManager.vFieldContiAct[1].X) / 2,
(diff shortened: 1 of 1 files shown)


DailyShana:
as there is no actual card in that place, the cards should be transparent.
and the place indicates conti act should be moved to the place above field zone:
image

mercury233:
the effect can be owned by any player, so we shouldn't make it look like only one player

mercury233:
it will be weird if multiple cards are drawn transparent

DailyShana:
> the effect can be owned by any player, so we shouldn't make it look like only one player

it's outside the field, so doesn't mean anything about where the effects come from

mercury233:
nobody will think a card above the field zone is outside the field

DailyShana:
> nobody will think a card above the field zone is outside the field

There is no frame. Contrarily, a card in the center of the field is more likely to be think as a card in the field

DailyShana:
> it will be weird if multiple cards are drawn transparent

only draw card image of the first card, make other cards completely transparent to indicate the count

mercury233:
image
image
image

more people prefer it in center, and it had been in center for a long time

PR #2319, opened 2020-09-17
add basic support for cant check grave

mercury233:
prevent checking grave when _Question_ is activating.

if card in grave must be selected, the order will be shuffled and the sequence won't be displayed.

lua modifation:
```lua
function c38723936.cost(e,tp,eg,ep,ev,re,r,rp,chk)
	if chk==0 then return true end
	if e:IsHasType(EFFECT_TYPE_ACTIVATE) then
		local e1=Effect.CreateEffect(e:GetHandler())
		e1:SetType(EFFECT_TYPE_FIELD)
		e1:SetProperty(EFFECT_FLAG_PLAYER_TARGET+EFFECT_FLAG_CLIENT_HINT)
		e1:SetDescription(CARD_QUESTION)
		e1:SetTargetRange(0,1)
		e1:SetReset(RESET_CHAIN)
		Duel.RegisterEffect(e1,tp)
	end
end
```

### DIFF
--- a/gframe/client_field.cpp
+++ b/gframe/client_field.cpp
@@ -29,6 +29,7 @@ ClientField::ClientField() {
 	conti_act = false;
 	deck_reversed = false;
 	conti_selecting = false;
+	cant_check_grave = false;
 	for(int p = 0; p < 2; ++p) {
 		mzone[p].resize(7, 0);
 		szone[p].resize(8, 0);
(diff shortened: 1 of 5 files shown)


mercury233:
image
The cards in grave will be shuffled, and the sequence won't be displayed.

The client can always be modified to record the grave changes, so we can never implement this card fully.

purerosefallen:
*There would be cheating clients everywhere.*
~~Finally I decided to ignore this packet in my fork~~

PR #2333, opened 2021-03-14
select bot deckfile

mercury233:
image


purerosefallen:
I suggest using param to pass the whole deck content. See my fork of windbot

PR #2337, opened 2021-03-26
fix ClientField::SetResponseSelectedOption()

salix5:
@mercury233 
Problem
Suppose that the trigger effect of (2nd) 驚楽園の支配人 ＜∀rlechino＞ is triggered 2 times in one chain.
After the chain ends, I can activate the effect of 驚楽園の支配人 ＜∀rlechino＞.

When I click the activate button of 驚楽園の支配人 ＜∀rlechino＞, there are 2 options since the 2nd effect is triggered twice.
That is, there are 2 chain object in core.select_chains, which are:
2nd effect(triggered by event group 1)
2nd effect(triggered by event group 2)

However, no matter which option I choose, I will always activate "2nd effect(triggered by event group 1)".

Reason
current ClientField::SetResponseSelectedOption()
The activatable_cards is a copy of core.select_chains in ocgcore.
It only compares the handler card and effect description when finding the response index.

In this case, the 2 chain object have the same triggering_effect, so the handler card and description are identical.
If the card and description are identical, it will always return the first chain object.

Solution
Require  Fluorohydride/ygopro-core#370
Now the index related to each option is stored in select_options_index.
ClientField::SetResponseSelectedOption() can find the correct index of the selected option.

Test replays
[replays.zip](https://github.com/Fluorohydride/ygopro/files/6212815/replays.zip)
bug_select.yrp
I choose the second option, but 2nd effect(triggered by event group 1) is activated.
correct_select.yrp
I choose the second option, and 2nd effect(triggered by event group 2) is activated as expected.

### DIFF
--- a/gframe/client_field.h
+++ b/gframe/client_field.h
@@ -42,6 +42,7 @@ public:
 	std::vector<ClientCard*> conti_cards;
 	std::vector<std::pair<int,int>> activatable_descs;
 	std::vector<int> select_options;
+	std::vector<int> select_options_index;
 	std::vector<ChainInfo> chains;
 	int extra_p_count[2];
 
--- a/gframe/event_handler.cpp
+++ b/gframe/event_handler.cpp
@@ -364,22 +364,23 @@ bool ClientField::OnEvent(const irr::SEvent& event) {
 				mainGame->wCmdMenu->setVisible(false);
 				ShowCancelOrFinishButton(0);
 				if(!list_command) {
-					int index = -1;
 					select_options.clear();
+					select_options_index.clear();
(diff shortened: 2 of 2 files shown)


mercury233:
may need a better way to handle the selection between 2 same effects

salix5:
The fist option: triggered by the first event group
The second option: triggered by the second event group

Although it is not perfect, my opinion is that it is acceptable as a workaround.

salix5:
@mercury233 
Can we merge this branch as a workaround?

mercury233:
confirming

salix5:
@mercury233 
Is there any problem in this branch?

PR #2346, opened 2021-05-19
add new path for arch linux cjk fonts

ygodueltech:
Earlier this year arch linux split CJK fonts into a separate package.
ygopro cashes on startup with no error for me without these added paths

------
**background on switch:**

https://bugs.archlinux.org/task/70196
https://archlinux.org/packages/extra/any/noto-fonts-cjk/


purerosefallen:
~~You should add fonts in every Linux distros~~

Issue #2352, opened 2021-06-06
Suggestion: make mtrandom.h not public


MonokuroInzanaito:
I have seen players reproducing algorithm in `mtrandom.h` when duel starts, making their starting hand always desirable.
To make an example, the player can check their starting hand at each second, and start the duel when the good hand comes. This breaks the fairness of the duels. I believe this is due to the publicness of `mtrandom.h`.
To make the duel fair enough, I think we should change the algorithm in `mtrandom.h`, and make that algorithm only available to developers.
An alternative solution, is to add a hidden key in this algorithm, or in the process seed generation. Both ways would make players reproducing the random algorithm impossible.

salix5:
@MonokuroInzanaito 
> I have seen players reproducing algorithm in `mtrandom.h` when duel starts, making their starting hand always desirable.
> To make an example, the player can check their starting hand at each second, and start the duel when the good hand comes. This breaks the fairness of the duels. I believe this is due to the publicness of `mtrandom.h`.

As far as I know, the reason is the seed of Random Number Generator(RNG).
Mersenne Twister is a verry popular RNG, and the algorithm is public all along.

ygopro used `time(nullp)` as the seed of RNG before, and the player can easily get it.
If a player can get the seed of RNG, he can duplicate the RNG and predict the starting hand.
This is a very inappropriate approach, and I try to fix it in #2362, #2363, Fluorohydride/ygopro-core#391.

> To make the duel fair enough, I think we should change the algorithm in `mtrandom.h`, and make that algorithm only available to developers.
> An alternative solution, is to add a hidden key in this algorithm, or in the process seed generation. Both ways would make players reproducing the random algorithm impossible.

Now ygopro will generate the seed by `std::random_device`, and I think the player cannot reproduce the seed anymore.
We appreciate your kind assistance, and please let us know if there is any problem in the new approach.

salix5:
@MonokuroInzanaito 
merged into master
If you find any problem, please give us your advice.
Thank you very much!

PR #2363, opened 2021-08-16
update RNG and uniform distribution, part 2

salix5:
@mercury233 
# Problem
1. The seed of RNG is the start time of the duel.
seed
These are 2 different games start at the same second.
All games start at the same second will have the same seeds because the unit of `time()` is second.
In other words, the starting hand is almost predictable.

2. The seed of `duel` is the first random number used in initial shuffle

Initial shuffle:
```c++
rnd.reset(seed);
for(size_t i = pdeck[0].main.size() - 1; i > 0; --i) {
	int swap = rnd.real() * (i + 1);
	std::swap(pdeck[0].main[i], pdeck[0].main[swap]);
}
```

Create duel:
```c++
rnd.reset(seed);
pduel = create_duel(rnd.rand());
```
Let {a_n} be the sequence generated by seed.

Initial shuffle:
a1, a2, a3, ...

duel seed:
a1

One can get some information about `a1` by observing `pdeck[0].main[n-1]`.

3. The "Shuffle" button in Deck Edit page uses `void random_shuffle( RandomIt first, RandomIt last )`
(In VS2019)
```c++
// FUNCTION TEMPLATE random_shuffle
template <class _RanIt>
void random_shuffle(_RanIt _First, _RanIt _Last) { // shuffle [_First, _Last) using rand()
    _Rand_urng_from_func _Func;
    _STD shuffle(_First, _Last, _Func);
}
```
`void random_shuffle( RandomIt first, RandomIt last )` is deprecated in c++14.
It should not be used anymore.

see  Fluorohydride/ygopro-core#391, #2352 

# Solution
Require #2362 

## Seed
Now the start time and seed are stored in different sections of the header.
The 2 seeds:
1. seed
Now it is generated by `std::random_device` rather than `time(nullptr)`.

2. duel_seed
Let {a_n} be the sequence generated by seed.

duel seed:
a1

Initial shuffle:
a2, a3, a4, ...,

## Shuffle
Now every shuffling uses `void mt19937::shuffle_vector(std::vector<T>& v)`.

### DIFF
--- a/gframe/config.h
+++ b/gframe/config.h
@@ -8,6 +8,7 @@
 #ifdef _WIN32
 
 #include <WinSock2.h>
+#define NOMINMAX
 #include <windows.h>
 #include <ws2tcpip.h>
 
--- a/gframe/deck_con.cpp
+++ b/gframe/deck_con.cpp
@@ -79,6 +79,7 @@ void DeckBuilder::Initialize() {
 	mainGame->btnSideReload->setVisible(false);
 	filterList = &deckManager._lfList[mainGame->gameConf.use_lflist ? mainGame->gameConf.default_lflist : deckManager._lfList.size() - 1].content;
 	ClearSearch();
+	rnd.reset((unsigned int)time(nullptr));
 	mouse_pos.set(0, 0);
 	hovered_code = 0;
 	hovered_pos = 0;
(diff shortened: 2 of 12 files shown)


purerosefallen:
https://github.com/purerosefallen/guess-hand
It reminded me about that...

salix5:
> https://github.com/purerosefallen/guess-hand
> It reminded me about that...
If you find any problem in the new version, please open a new issue.
Thank you in advance.

PR #2368, opened 2021-08-30
update Simplified Chinese ot rule

mercury233:
Alternative to https://github.com/Fluorohydride/ygopro/pull/2358

-----

image
image
image
image


salix5:
## Suggestions:
### Add DECKERROR_SCONLY in network.h
I think we should give a deck error code to each language.
The length of card id: 28 bit
The length of deck error msg: 4 bit, 0x1~0xf
We can still add new deck error code.

### Move the beginning of system string to 1480
`host.rule` is 0-based, and the range is 0 to 3 (now) or 0 to 5 (new).
If 1480 is still available, I think putting them at 1480~1485 is more intuitive.


## Questions:
### ot
The new `ot` column of cards.cdb will be a bit field:
```c++
#define AVAIL_OCG					0x1
#define AVAIL_TCG					0x2
#define AVAIL_OCGTCG				0x3
#define AVAIL_CUSTOM				0x4
#define AVAIL_SC					0x8
```
Is it correct?

mercury233:
@salix5 Added the general hint
I prefer to preserve `1480` for future using
The `ot` column is already in the database as integer

salix5 (approved):

gframe/drawing.cpp
```diff
@@ -1235,11 +1238,11 @@ void Game::DrawDeckBd() {
 			const wchar_t* ptype = dataManager.FormatType(ptr->second.type);
 			DrawShadowText(textFont, ptype, Resize(860, 187 + i * 66, 955, 207 + i * 66), Resize(1, 1, 0, 0));
 			textBuffer[0] = 0;
-			if((ptr->second.ot & 0x3) == 1)
+			if((ptr->second.ot & AVAIL_OCGTCG) == AVAIL_OCG)
 				wcscat(textBuffer, L"[OCG]");
-			else if((ptr->second.ot & 0x3) == 2)
+			else if((ptr->second.ot & AVAIL_OCGTCG) == AVAIL_TCG)
 				wcscat(textBuffer, L"[TCG]");
```
purerosefallen, line 1244: how about `AVAIL_SC` here?

gframe/drawing.cpp
```diff
@@ -1105,23 +1105,26 @@ void Game::DrawThumb(code_pointer cp, position2di pos, const std::unordered_map<
...
+		bool avail = !((filter_lm == 4 && !(cp->second.ot & AVAIL_OCG)
+					|| (filter_lm == 5 && !(cp->second.ot & AVAIL_TCG))
+					|| (filter_lm == 6 && !(cp->second.ot & AVAIL_SC))
+					|| (filter_lm == 7 && !(cp->second.ot & AVAIL_CUSTOM))
+					|| (filter_lm == 8 && (cp->second.ot & AVAIL_OCGTCG) != AVAIL_OCGTCG)));
+		if(avail) {
+			if((cp->second.ot & AVAIL_OCG) && !(cp->second.ot & AVAIL_TCG))
+				driver->draw2DImage(imageManager.tOT, otloc, recti(0, 128, 128, 192), 0, 0, true);
+			else if((cp->second.ot & AVAIL_TCG) && !(cp->second.ot & AVAIL_OCG))
+				driver->draw2DImage(imageManager.tOT, otloc, recti(0, 192, 128, 256), 0, 0, true);
```
purerosefallen, line 1119: same here. I think we need to add a texture of SC hover.

mercury233:
maybe we should add a switch to display SC release status of cards, _in future_

PR #2372, opened 2021-09-05
check the filename generated from the player name

mercury233:
image

will be saved to

image


mercury233:
Should we make those conversion in the deck and replay file name entered by the user himself?

PR #2399, opened 2021-11-29
update build system

mercury233:
https://github.com/Fluorohydride/ygopro-core/pull/417


purerosefallen:
Better to support macOS universal build.

PR #2405, opened 2021-12-07
add search compare attack/defense

465uytrewq:
user can search ATK equal DEF similar to Yu-Gi-Oh Neuron (for Gizmek and Spirit Sculptor).

need these in strings.conf:

```
!system 1375 Compare ATK/DEF
!system 1376 Attack
!system 1377 Defense
```

search001search002search003


mercury233:
In my option, the GUI of a duel simulator shouldn't contains such complex feature, which should be done by expert card search tools.

PR #2414, opened 2022-01-11
Add initial support for build.zig

sagehane:
Proposal
---
I was wondering if this project would be willing to look into the [Zig build system](https://ziglang.org/learn/overview/#zig-build-system).

The idea is that, theoretically, a person can install Zig and the relevant dependencies and build this project for [any of the target supported by Zig](https://ziglang.org/download/0.9.0/release-notes.html#Support-Table).

It also enables stuff such as UBSAN by default, which seems to have spotted some undefined behaviors for this project already.

TODO:
---

- [ ] Get the build to work on all relevant platforms. (Currently only supports `x86_64-linux`)
- [ ] Resolve what to do with `irrlicht` headers. 
I renamed `irrlicht.h` -> `irrlicht/irrlicht.h` for example. Curious why it currently is that way.
- [ ] Get the build working on Debian-based distros that seem to expect `lua-c++` instead of `lua`.
- [ ] Figure out how to handle the files described in https://github.com/Fluorohydride/ygopro/wiki#running.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+# Zig files
+zig-cache/
+zig-out/
--- a/build.zig
+++ b/build.zig
@@ -1,0 +1,115 @@
+const std = @import("std");
+
+pub fn build(b: *std.build.Builder) void {
+    const target = b.standardTargetOptions(.{});
+    const mode = b.standardReleaseOptions();
+
+    const c_flags = [_][]const u8{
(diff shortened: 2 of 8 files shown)


mercury233:
Zig seems cool but I'm afraid I can't provide too much help.

For Irrlicht, the release from Irrlicht official organize files like `irrlicht/include/*.h` and `irrlicht/source/Irrlicht/*.cpp` so we added `irrlicht/include` to include in premake.

PR #2452, opened 2022-09-01
Support multple lan clients from the same IP

edo9300:
Currently if 2 clients host a lan room on a single machine, only one of the 2 rooms will be able to answer to broadcast messages, use ``SO_REUSEADDR`` so that multiple sockets can be bound to the same port (7920 in this case).
Also update the duelclient part to no longer filter the existing answer based off the ip but also on the port.

### DIFF
--- a/gframe/duelclient.cpp
+++ b/gframe/duelclient.cpp
@@ -4031,9 +4031,9 @@ void DuelClient::BroadcastReply(evutil_socket_t fd, short events, void * arg) {
 		socklen_t sz = sizeof(sockaddr_in);
 		char buf[256];
 		/*int ret = */recvfrom(fd, buf, 256, 0, (sockaddr*)&bc_addr, &sz);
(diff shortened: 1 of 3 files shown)


gframe/duelclient.cpp
```diff
@@ -4031,11 +4031,16 @@ void DuelClient::BroadcastReply(evutil_socket_t fd, short events, void * arg) {
 		socklen_t sz = sizeof(sockaddr_in);
 		char buf[256];
 		/*int ret = */recvfrom(fd, buf, 256, 0, (sockaddr*)&bc_addr, &sz);
-		unsigned int ipaddr = bc_addr.sin_addr.s_addr;
+		if(is_closing || pHP->identifier == NETWORK_SERVER_ID)
```
purerosefallen, line 4034: `pHP` comes from nowhere, it can't compile.

purerosefallen (changes requested):

PR #2457, opened 2022-09-20
add SRV record support

purerosefallen:
image

~~The port input box would be removed, and server port is typed in after `:`~~
The host field supports SRV record parsing, so that players can join servers without a port. The following is an example of `SRV` record, in which players could join with hostname `koishi.momobako.com` and it would make the client connect to `koishi.momobako.com:7210`

In addition, now it also supports format of `koishi.momobako.com:7210` with port field blank.

```
➜  ~ dig _ygopro._tcp.koishi.momobako.com srv                           14:22:46

; <<>> DiG 9.18.1-1ubuntu1.1-Ubuntu <<>> _ygopro._tcp.koishi.momobako.com srv
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 7031
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 2

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;_ygopro._tcp.koishi.momobako.com. IN	SRV

;; ANSWER SECTION:
_ygopro._tcp.koishi.momobako.com. 60 IN	SRV	0 5 7210 koishi.momobako.com.

;; ADDITIONAL SECTION:
koishi.momobako.com.	60	IN	A	10.198.6.95

;; Query time: 23 msec
;; SERVER: 10.198.7.56#53(10.198.7.56) (UDP)
;; WHEN: Tue Sep 20 14:32:19 CST 2022
;; MSG SIZE  rcvd: 167

```


mercury233:
I prefer to keep the port inputbox but accept blank value

purerosefallen:
> I prefer to keep the port inputbox but accept blank value

@mercury233 Players would still have to change text in 2 boxes for switching servers.
After this one, servers should be recognized as a single address, either with `address:port` format, or single address for SRV record.

purerosefallen:
Will redo one later for too many conflicts

PR #2458, opened 2022-09-22
fix Game::DrawGUI()

salix5:
@mercury233 

# Problem
bug

Replay:
[bug_texture.zip](https://github.com/Fluorohydride/ygopro/files/9622419/bug_texture.zip)

Fusion Summon ティアラメンツ・キトカロス
Link Summon I：Pマスカレーナ
Activated the 2nd effect of ティアラメンツ・キトカロス (target: ティアラメンツ・キトカロス)
Special Summon ティアラメンツ・メイルゥ 
Selecting position (The bug occurs.)

# Reason
The image on the defense position button is ティアラメンツ・キトカロス.
It is probably because some buttons in `imageLoading` are unprocessed.

# Solution
Remove all buttons in `imageLoading`.

### DIFF
--- a/gframe/drawing.cpp
+++ b/gframe/drawing.cpp
@@ -685,10 +685,10 @@ void Game::DrawStatus(ClientCard* pcard, int x1, int y1, int x2, int y2) {
 	}
 }
 void Game::DrawGUI() {
-	if(imageLoading.size()) {
-		for(auto mit = imageLoading.begin(); mit != imageLoading.end(); ++mit)
-			mit->first->setImage(imageManager.GetTexture(mit->second));
-		imageLoading.clear();
+	while (imageLoading.size()) {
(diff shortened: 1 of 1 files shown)


mercury233:
Can you reproduce that situation? I can't, need more detail.

purerosefallen:
@mercury233 this is because on slow computers, image load queue is filled within loading process. It's was a bad design on image loading queue, so I believe this fix looks good.

salix5:
> Can you reproduce that situation? I can't, need more detail.

No, the player who upload that replay file has no more detail.

However, it had happened before.
unknown

My opinion:
`imageLoading` is a to do list, so all buttons in the list must be taken out.

If a button is added into `imageLoading` after the for loop begins, the button can be omitted sometimes.
Therefore, I think we should change this part in order to make sure that all buttons are taken out.

PR #2468, opened 2022-11-06
update MSG_MOVE for xyz deattach

mercury233:


purerosefallen:
Still useful? Perhaps we could make them to be 0x1362

PR #2470, opened 2022-11-12
update MSG_CONFIRM_CARDS, add skip_panel

mercury233:
GIF 2022-11-12 10-24-31
When using the effect of _Number 89_ or Dogmatika cards, the player will expect the card to be selectable on confirm, but now we give the player 2 dialog, in the first one the player can only look at the cards, in the second one the player can select.
We can add `skip_panel` to skip the first dialog, let the `MSG_CONFIRM_CARDS` only do the update code data job.


purerosefallen:
Please open a PR to SRVPro and WindBot about this change.

PR #2494, opened 2023-05-03
add RACE_ILLUSION, RACES_COUNT

mercury233:
>Duelist Nexus introduces the first new monster Type to the Yu-Gi-Oh! TRADING CARD GAME since Cyberse monsters were introduced in 2017’s Starter Deck: Link Strike! **Illusion** monsters cannot be struck down in battle. They can’t defeat other monsters in battle either, but they harm them in other ways instead! Check out Nightmare Magician, an Illusion monster that gives you control of any monster it battles! It can also destroy any other card on the field if another monster attacks!
https://www.yugioh-card.com/en/products/dune/

### DIFF
--- a/gframe/data_manager.cpp
+++ b/gframe/data_manager.cpp
@@ -279,7 +279,7 @@ const wchar_t* DataManager::FormatRace(int race) {
 	wchar_t* p = racBuffer;
 	unsigned filter = 1;
 	int i = 1020;
-	for(; filter != 0x2000000; filter <<= 1, ++i) {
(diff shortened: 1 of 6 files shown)


rumia-san:
> [Fluorohydride/ygopro-core#495](https://github.com/Fluorohydride/ygopro-core/pull/495) https://github.com/Fluorohydride/ygopro-core/blob/61bf5a9b598295d9cdd41cbd1f11267738afbc07/common.h#L110

谢谢！
（这个repo的submodule版本是不是应该更新一下了）

salix5:
ygopro-core可以單獨更新

PR #2501, opened 2023-08-13
remove usage of std::unary_function in irrUString.h

mercury233:
https://en.cppreference.com/w/cpp/utility/functional/unary_function
https://stackoverflow.com/questions/63577103/what-is-an-equivalent-replacement-for-stdunary-function-in-c17

It is deprecated in C++11 and removed in C++17

With this code, YGOPro can't build in VS2022 17.7.0. The error is C2039 `'unary_function' : is not a member of 'std'`

Even through I doubt it is a bug because I had `/std:c++14` on, this code seems can be removed.

### DIFF
--- a/gframe/irrUString.h
+++ b/gframe/irrUString.h
@@ -3052,7 +3052,7 @@ namespace unicode {
 
 //! Hashing algorithm for hashing a ustring.  Used for things like unordered_maps.
 //! Algorithm taken from std::hash<std::string>.
-class hash : public std::unary_function<core::ustring, size_t> {
+class hash {
 public:
 	size_t operator()(const core::ustring& s) const {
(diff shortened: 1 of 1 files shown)


salix5:
Files from suckerfreegames:
CGUITTFont.h
CGUITTFont.cpp
irrUString.h

https://github.com/Fluorohydride/ygopro/commits/master/gframe/irrUString.h
The file in ygopro is added in 2011

https://irrlicht.suckerfreegames.com/i18n/base/
The file in suckerfreegames
Last update: 2017/3/1

My conclusion:
unicode::hash() is not used in this project, so we can safely modify or remove it.
The last version is released 6 years ago.
Even if we want to use hash in the future, we should use std::hash() instead.

PR #2519, opened 2024-03-22
move OnCommonEvent to class Game 

salix5:
@mercury233 @purerosefallen 
```cpp
class Game{
	ClientField dField;
	DeckBuilder deckBuilder;
	MenuHandler menuHandler;
}

class ClientField: public irr::IEventReceiver
class DeckBuilder: public irr::IEventReceiver
class MenuHandler: public irr::IEventReceiver
```
`OnCommonEvent` seems to be the event handler for common events of these `IEventReceiver`.
Maybe it should be put in class Game.


mercury233:
looks not pretty

PR #2520, opened 2024-03-24
change cbCardType2 width

salix5:
@mercury233  @purerosefallen 
The width of cbCardType2
long ago: 75
now: 60

The width is too small for some types.
Now it is changed into 70.


mercury233:
I think using magic variable names is more confusing than using numbers (magic or not)

salix5:
it is not a magical variable name.
It is literally
X1 (left border) of combo box cbCardType2
X2 (right border) of combo box cbCardType2

It should be a fixed number in declaration and resize, but now it is not.

PR #2524, opened 2024-04-06
use memcpy in bufferio.h

salix5:
Require:
https://github.com/Fluorohydride/ygopro-core/pull/576

https://gist.github.com/shafik/848ae25ee209f698763cffee272a58f8
According to the strict aliasing rule, we should not cast a pointer should to another type and dereference.
It can be done by `memcpy`.


salix5:
> well, `unsigned char` type is allowed, so there is no problem here.
> 
> > What the C++17 Draft Standard say
> > If a program attempts to access the stored value of an object through a glvalue of **other than** one of the following types the behavior is undefined:
> > ...
> > (11.8) — a char, unsigned char, or std::byte type.
> 
> but the code still can be improved, we can change C-style cast to C++ `static_cast`

I think it means that:
A. cast `int*` to `unsigned char*`, access by `unsigned char*`: OK
B. cast `unsigned char*` to `int*`, access by `int*`: not OK

ygopro is type B, which breaks the strict aliasing rule.

DailyShana:
You are right, there was an explanation long ago I forgot 😂: https://github.com/Fluorohydride/ygopro-core/pull/176#issuecomment-412242659
we can try to remove flag `-fno-strict-aliasing` 

and it's better to have an `std::` because `<cxxx>` headers definitely provide `std::xxxx` but may not provide `xxxx`

salix5:
@mercury233 @purerosefallen 
Can we merge this one?

PR #2525, opened 2024-04-07
remove unnecessary pointer cast

salix5:
@mercury233 
struct sqlite3_vfs_t
user-defined type

https://github.com/Fluorohydride/ygopro/blob/4c0ca0f1b9b8c9b0389ca43f67bc8070801509fd/gframe/spmemvfs/spmemvfs.c#L259
There is only 1 `sqlite3_vfs_t` object, g_spmemvfs, so there is only 1 `sqlite3_vfs` object, g_spmemvfs.base.

The pointer cast breaks strict aliasing rule, and it is unnecessary.

### DIFF
--- a/gframe/spmemvfs/spmemvfs.c
+++ b/gframe/spmemvfs/spmemvfs.c
@@ -239,7 +239,6 @@ typedef struct spmemvfs_cb_t {
 } spmemvfs_cb_t;
 
 typedef struct spmemvfs_t {
-	sqlite3_vfs base;
 	spmemvfs_cb_t cb;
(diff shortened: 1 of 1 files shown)


DailyShana:
A pointer to a struct can be cast to a pointer to its first member and vice versa. This do not violate strict aliasing rule
https://stackoverflow.com/questions/50383187/is-it-a-strict-aliasing-violation-to-alias-a-struct-as-its-first-member
https://stackoverflow.com/questions/60536422/pointer-aliasing-between-struct-and-first-member-of-struct

PR #2528, opened 2024-04-18
DataManager: add MentionHandler

salix5:
@mercury233 @purerosefallen 
Require:
https://github.com/Fluorohydride/ygopro-core/pull/579

The mention handler implementation.

Add 2 System strings:
```
!system 50 (Left border of the card name)
!system 51 (Right border of the card name)
```
The default values are `"` and `"`.


purerosefallen:
I don't suggest you to waste more time on this PR, until a proper solution to the server deployment and the problem of server memory cost come out.

PR #2533, opened 2024-04-22
update client_card_sort()

salix5:


mercury233:
It had checked `c1->location != c2->location` before, therefore `c2` must also be in `LOCATION_OVERLAY`, so `c2` should have overlay target.

salix5:
The pointer should be checked before dereferencing.

DailyShana:
if `c1->location` has `LOCATION_OVERLAY` but `c1->overlay_target` is empty, there must be something wrong elsewhere and in this case checking will hide the problem.
Some IDE that automatically analysis the code quality will give a warning because it do not understand this underlying logic.

PR #2540, opened 2024-05-02
fix EncodeUTF8,  DecodeUTF8

salix5:
# Problem
EncodeUTF8
DecodeUTF8
These functions did not check the size of output array.
If the source string is too long, it will pass the boundary of the output array.

Example:
Create a file in `./single`
```
五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二一五四三二.lua
```
The filename is about 692 bytes in UTF-8.
The process will crash when trying to open this file.

# Solution
- The macros can be replaced by sizeof(wchar_t) 
- The size is deduced by the template
- Convert char/wchar_t to unsigned before bitwise operation
- Replace `sprintf` with `snprintf`

# Reference
https://github.com/edo9300/edopro/blob/master/gframe/bufferio.h

@edo9300 
My revision:
C++11
4.5 Integral promotions 
> A prvalue of an integer type other than bool, char16_t, char32_t, or wchar_t whose integer conversion rank (4.13) is less than the rank of int can be converted to a prvalue of type int if int can represent all the values of the source type; otherwise, the source prvalue can be converted to a prvalue of type unsigned int.

```cpp
char x=0xff;
x<<16; //promoted to int
```
The operand of bitwise operator should be unsigned type, so I think char/wchar_t should be cast to unsigned int first.

@mercury233 
@purerosefallen

### DIFF
--- a/gframe/bufferio.h
+++ b/gframe/bufferio.h
@@ -61,34 +61,31 @@ public:
 	static int EncodeUTF8(const wchar_t * wsrc, char * str) {
 		char* pstr = str;
 		while(*wsrc != 0) {
-			if(*wsrc < 0x80) {
-				*str = (char)*wsrc;
+			unsigned cur = *wsrc;
+			if(cur < 0x80) {
+				*str = (char)cur;
 				++str;
-			} else if(*wsrc < 0x800) {
-				str[0] = ((*wsrc >> 6) & 0x1f) | 0xc0;
-				str[1] = ((*wsrc) & 0x3f) | 0x80;
+			} else if(cur < 0x800) {
+				str[0] = ((cur >> 6) & 0x1f) | 0xc0;
+				str[1] = (cur & 0x3f) | 0x80;
 				str += 2;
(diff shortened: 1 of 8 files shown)


salix5:
@mercury233 
@purerosefallen 
Can we merge this one?

purerosefallen:
Could you do a merge to mycard/ygopro:server?

salix5:
https://github.com/mycard/ygopro/compare/server...Fluorohydride:ygopro:master
It seems that it is solved.

PR #2547, opened 2024-05-13
add 0x3 Allure Queen

salix5:
@mercury233 
@purerosefallen


mercury233:
I guess 0x3 was Horus, but anyway it don't matter now.

PR #2551, opened 2024-05-21
 disconnect when the packet is too large 

### LINKED ISSUE #2314: Security Issue: Memory leak.
#### Memory leak

This vulnerability happened when the function "SingleDuel::UpdateDeck" and "SingleDuel::PlayerReady" been used.
When the player sends a package with error 'mainc' size and 'sidec' size, the function "SingleDuel::UpdateDeck" haven't check those parameters is legal or not. Then this function will calculate the sum of those parameters.

```c
void SingleDuel::UpdateDeck(DuelPlayer* dp, void* pdata, unsigned int len) {
	if(dp->type > 1 || ready[dp->type])
		return;
	char* deckbuf = (char*)pdata;
	int mainc = BufferIO::ReadInt32(deckbuf);		// main_len
	int sidec = BufferIO::ReadInt32(deckbuf);		// side_len
	// verify data
	if((unsigned)mainc + (unsigned)sidec > (len - 8) / 4) {
		STOC_ErrorMsg scem;
		scem.msg = ERRMSG_DECKERROR;
		scem.code = 0;
		NetServer::SendPacketToPlayer(dp, STOC_ERROR_MSG, scem);
		return;
	}
	if(duel_count == 0) {
		deck_error[dp->type] = deckManager.LoadDeck(pdeck[dp->type], (int*)deckbuf, mainc, sidec);
	} else {
		if(deckManager.LoadSide(pdeck[dp->type], (int*)deckbuf, mainc, sidec)) {
			ready[dp->type] = true;
			NetServer::SendPacketToPlayer(dp, STOC_DUEL_START);
			if(ready[0] && ready[1]) {
				NetServer::SendPacketToPlayer(players[tp_player], STOC_SELECT_TP);
				players[1 - tp_player]->state = 0xff;
				players[tp_player]->state = CTOS_TP_RESULT;
				duel_stage = DUEL_STAGE_FIRSTGO;
			}
		} else {
			STOC_ErrorMsg scem;
			scem.msg = ERRMSG_SIDEERROR;
			scem.code = 0;
			NetServer::SendPacketToPlayer(dp, STOC_ERROR_MSG, scem);
		}
	}
}
```

The algorithm thinks 'mainc' and 'sidec' are two unsigned number, so there is an integer overflow, when we set those parameters like (1, -1), their sum will be zero. It is ok though but in the function "deckManager.LoadDeck", it will cause a buffer overread.

```c
int DeckManager::LoadDeck(Deck& deck, int* dbuf, int mainc, int sidec) {
	deck.clear();
	int code;
	int errorcode = 0;
	CardData cd;
	for(int i = 0; i < mainc; ++i) {
		code = dbuf[i];
		if(!dataManager.GetData(code, &cd)) {
			errorcode = code;
			continue;
		}
		if(cd.type & TYPE_TOKEN)
			continue;
		else if(cd.type & (TYPE_FUSION | TYPE_SYNCHRO | TYPE_XYZ | TYPE_LINK) && deck.extra.size() < 15) {
			deck.extra.push_back(dataManager.GetCodePointer(code));	//verified by GetData()
		} else if(deck.main.size() < 60) {
			deck.main.push_back(dataManager.GetCodePointer(code));
		}
	}
	for(int i = 0; i < sidec; ++i) {
		code = dbuf[mainc + i];
		if(!dataManager.GetData(code, &cd)) {
			errorcode = code;
			continue;
		}
		if(cd.type & TYPE_TOKEN)
			continue;
		if(deck.side.size() < 15)
			deck.side.push_back(dataManager.GetCodePointer(code));	//verified by GetData()
	}
	return errorcode;
}
```

We can see in this function, the parameters 'mainc' and 'sidec' were treat as two int type numbers. So if we set 'mainc' as '0x7fffffff' and 'sidec' as '0x80000001', in this function it will read 'dbuf' from range 0 to 2147483647.

Function "DeckManager::LoadDeck" also do a important thing: If the memory's data can be found in "dataManager.GetData" function, then it will add it into 'deck'; Else if "dataManager.GetData" function can't find memory's data in database, it will record it in the errorcode and return. Then in the "SingleDuel::UpdateDeck" function, the errorcode will be writen into 'deck_error[dp->type]'.

Now we can see the function "SingleDuel::PlayerReady".

```c
void SingleDuel::PlayerReady(DuelPlayer* dp, bool is_ready) {
	if(dp->type > 1)
		return;
	if(ready[dp->type] == is_ready)
		return;
	if(is_ready) {
		unsigned int deckerror = 0;
		if(!host_info.no_check_deck) {
			if(deck_error[dp->type]) {
				deckerror = (DECKERROR_UNKNOWNCARD << 28) + deck_error[dp->type];
			} else {
				bool allow_ocg = host_info.rule == 0 || host_info.rule == 2;
				bool allow_tcg = host_info.rule == 1 || host_info.rule == 2;
				deckerror = deckManager.CheckDeck(pdeck[dp->type], host_info.lflist, allow_ocg, allow_tcg);
			}
		}
		if(deckerror) {
			STOC_HS_PlayerChange scpc;
			scpc.status = (dp->type << 4) | PLAYERCHANGE_NOTREADY;
			NetServer::SendPacketToPlayer(dp, STOC_HS_PLAYER_CHANGE, scpc);
			STOC_ErrorMsg scem;
			scem.msg = ERRMSG_DECKERROR;
			scem.code = deckerror;
			NetServer::SendPacketToPlayer(dp, STOC_ERROR_MSG, scem);
			return;
		}
	}
	ready[dp->type] = is_ready;
	STOC_HS_PlayerChange scpc;
	scpc.status = (dp->type << 4) | (is_ready ? PLAYERCHANGE_READY : PLAYERCHANGE_NOTREADY);
	NetServer::SendPacketToPlayer(players[dp->type], STOC_HS_PLAYER_CHANGE, scpc);
	if(players[1 - dp->type])
		NetServer::SendPacketToPlayer(players[1 - dp->type], STOC_HS_PLAYER_CHANGE, scpc);
	for(auto pit = observers.begin(); pit != observers.end(); ++pit)
		NetServer::SendPacketToPlayer(*pit, STOC_HS_PLAYER_CHANGE, scpc);
#ifdef YGOPRO_SERVER_MODE
	if(cache_recorder)
		NetServer::SendPacketToPlayer(cache_recorder, STOC_HS_PLAYER_CHANGE, scpc);
	if(replay_recorder)
		NetServer::SendPacketToPlayer(replay_recorder, STOC_HS_PLAYER_CHANGE, scpc);
#endif
}
```

This function will check 'deck_error[dp->type]', if it is not zero, it will pack the errcode into 'scem' structure and send it to players. Thanks to it, we can get an easy way to leak the program memory.

And here is my test program, it works well in my local environment.

https://github.com/ChinaBluecat/ygo-fuzz-frame

salix5:
-  use memcpy in DuelClient, NetServer 
-  replace memcpy with std::memcpy 

# Buffer overflow in current version
9570e838a0afc9630e478bb456656f5207708f59
Before:
The server will copy `packet_len + 2` bytes to buffer, where `packet_len` is sent by the client.
It will suffer from buffer overflow if the client sends a very large packet.
(crash and/or security issues)

Now:
The server will disconnect when the packet is too large .

e8209d5d38a6a3f8d8274d0b41be071bf006c32e
8d542df6cbe0856a439cf3528afd8171f78cdbae
Before:
The server did not check the length of current data before handling the current packet.
The server might read data from previous packet.

Now:
The server will check the packet before handling.

# Remove `struct STOC_Chat`
Before:
```cpp
struct STOC_Chat {
	unsigned short player;
	unsigned short msg[256];
};
```
The size of STOC_Chat: 2+512=514
But the length of almost every `msg` sent by server is less than 256.

If the length is 100, the buffer would be:
player: 2 bytes
msg: 100*2=200 bytes
(total: 202)

If the client treat this buffer as a `struct STOC_Chat`, it will go out of bounds.
It indicates that the buffer should be treated as a `uint_16` array, and `struct STOC_Chat` is unnecessary.

Now:
The buffer is treated as a `uint_16` array.
The buffer size is checked by 
```cpp
bool check_msg_size(int size);
```

# `bool` in `struct HostInfo` is replaced by `unsigned char`
https://en.cppreference.com/w/cpp/language/types
> bool — integer type, capable of holding one of the two values: true or false. The value of sizeof(bool) is implementation defined and might differ from 1. 

If we use struct to exchange data, the size of every member should be fixed.

# fix DuelMode::UpdateDeck() 
Before: 
```cpp
const unsigned int possibleMaxLength = (len - 8) / 4;
```
If len = 4U
possibleMaxLength =1073741823U
The check will fail.

Now:
Use signed integer compare
Check the value, alignment of `len`
Check the value of `mianc`, `sidec`
Check the value of `deck_size` 
fix #2314 

# Reference
edo9300/edopro@533b300

@mercury233 
@purerosefallen

### DIFF
--- a/gframe/deck_con.cpp
+++ b/gframe/deck_con.cpp
@@ -1757,7 +1757,7 @@ bool DeckBuilder::push_main(code_pointer pointer, int seq) {
 	if(pointer->second.type & (TYPE_FUSION | TYPE_SYNCHRO | TYPE_XYZ | TYPE_LINK))
 		return false;
 	auto& container = deckManager.current_deck.main;
-	int maxc = mainGame->is_siding ? 64 : DECK_MAX_SIZE;
+	int maxc = mainGame->is_siding ? DECK_MAX_SIZE + 4 : DECK_MAX_SIZE;
 	if((int)container.size() >= maxc)
 		return false;
 	if(seq >= 0 && seq < (int)container.size())
@@ -1772,7 +1772,7 @@ bool DeckBuilder::push_extra(code_pointer pointer, int seq) {
 	if(!(pointer->second.type & (TYPE_FUSION | TYPE_SYNCHRO | TYPE_XYZ | TYPE_LINK)))
 		return false;
 	auto& container = deckManager.current_deck.extra;
-	int maxc = mainGame->is_siding ? 20 : EXTRA_MAX_SIZE;
+	int maxc = mainGame->is_siding ? EXTRA_MAX_SIZE + 5 : EXTRA_MAX_SIZE;
 	if((int)container.size() >= maxc)
 		return false;
 	if(seq >= 0 && seq < (int)container.size())
@@ -1785,7 +1785,7 @@ bool DeckBuilder::push_extra(code_pointer pointer, int seq) {
 }
 bool DeckBuilder::push_side(code_pointer pointer, int seq) {
 	auto& container = deckManager.current_deck.side;
-	int maxc = mainGame->is_siding ? 20 : SIDE_MAX_SIZE;
+	int maxc = mainGame->is_siding ? SIDE_MAX_SIZE + 5 : SIDE_MAX_SIZE;
 	if((int)container.size() >= maxc)
 		return false;
 	if(seq >= 0 && seq < (int)container.size())
--- a/gframe/deck_con.cpp
+++ b/gframe/deck_con.cpp
@@ -1249,7 +1249,7 @@ void DeckBuilder::GetHoveredCard() {
 		} else if(y >= 164 && y <= 435) {
 			int lx = 10, px, py = (y - 164) / 68;
 			hovered_pos = 1;
-			if(deckManager.current_deck.main.size() > 40)
+			if(deckManager.current_deck.main.size() > DECK_MIN_SIZE)
 				lx = (deckManager.current_deck.main.size() - 41) / 4 + 11;
 			if(x >= 750)
 				px = lx - 1;
@@ -1757,7 +1757,7 @@ bool DeckBuilder::push_main(code_pointer pointer, int seq) {
 	if(pointer->second.type & (TYPE_FUSION | TYPE_SYNCHRO | TYPE_XYZ | TYPE_LINK))
 		return false;
 	auto& container = deckManager.current_deck.main;
-	int maxc = mainGame->is_siding ? 64 : 60;
+	int maxc = mainGame->is_siding ? 64 : DECK_MAX_SIZE;
 	if((int)container.size() >= maxc)
 		return false;
 	if(seq >= 0 && seq < (int)container.size())
@@ -1772,7 +1772,7 @@ bool DeckBuilder::push_extra(code_pointer pointer, int seq) {
 	if(!(pointer->second.type & (TYPE_FUSION | TYPE_SYNCHRO | TYPE_XYZ | TYPE_LINK)))
 		return false;
 	auto& container = deckManager.current_deck.extra;
-	int maxc = mainGame->is_siding ? 20 : 15;
+	int maxc = mainGame->is_siding ? 20 : EXTRA_MAX_SIZE;
 	if((int)container.size() >= maxc)
 		return false;
 	if(seq >= 0 && seq < (int)container.size())
@@ -1785,7 +1785,7 @@ bool DeckBuilder::push_extra(code_pointer pointer, int seq) {
 }
 bool DeckBuilder::push_side(code_pointer pointer, int seq) {
 	auto& container = deckManager.current_deck.side;
-	int maxc = mainGame->is_siding ? 20 : 15;
+	int maxc = mainGame->is_siding ? 20 : SIDE_MAX_SIZE;
 	if((int)container.size() >= maxc)
 		return false;
 	if(seq >= 0 && seq < (int)container.size())
--- a/gframe/data_manager.cpp
+++ b/gframe/data_manager.cpp
@@ -61,7 +61,7 @@ bool DataManager::LoadDB(const wchar_t* wfile) {
 					if (len > SIZE_SETCODE)
 						len = SIZE_SETCODE;
 					if (len)
-						memcpy(cd.setcode, it->second.data(), len * sizeof(uint16_t));
+						std::memcpy(cd.setcode, it->second.data(), len * sizeof(uint16_t));
 				}
 				else
 					cd.set_setcode(setcode);
@@ -184,7 +184,7 @@ bool DataManager::GetData(unsigned int code, CardData* pData) {
 	if (pData) {
 		pData->code = data.code;
 		pData->alias = data.alias;
-		memcpy(pData->setcode, data.setcode, SIZE_SETCODE);
+		std::memcpy(pData->setcode, data.setcode, SIZE_SETCODE);
 		pData->type = data.type;
 		pData->level = data.level;
 		pData->attribute = data.attribute;
--- a/gframe/deck_manager.cpp
+++ b/gframe/deck_manager.cpp
@@ -86,11 +86,11 @@ int DeckManager::CheckDeck(Deck& deck, int lfhash, int rule) {
 	if(!list)
 		return 0;
 	int dc = 0;
-	if(deck.main.size() < 40 || deck.main.size() > 60)
+	if(deck.main.size() < DECK_MIN_SIZE || deck.main.size() > DECK_MAX_SIZE)
 		return (DECKERROR_MAINCOUNT << 28) + deck.main.size();
-	if(deck.extra.size() > 15)
+	if(deck.extra.size() > EXTRA_MAX_SIZE)
 		return (DECKERROR_EXTRACOUNT << 28) + deck.extra.size();
-	if(deck.side.size() > 15)
+	if(deck.side.size() > SIDE_MAX_SIZE)
 		return (DECKERROR_SIDECOUNT << 28) + deck.side.size();
 	const int rule_map[6] = { AVAIL_OCG, AVAIL_TCG, AVAIL_SC, AVAIL_CUSTOM, AVAIL_OCGTCG, 0 };
 	int avail = rule_map[rule];
@@ -158,10 +158,10 @@ int DeckManager::LoadDeck(Deck& deck, int* dbuf, int mainc, int sidec, bool is_p
 			continue;
 		}
 		else if(cd.type & (TYPE_FUSION | TYPE_SYNCHRO | TYPE_XYZ | TYPE_LINK)) {
-			if(deck.extra.size() >= 15)
+			if(deck.extra.size() >= EXTRA_MAX_SIZE)
 				continue;
 			deck.extra.push_back(dataManager.GetCodePointer(code));
-		} else if(deck.main.size() < 60) {
+		} else if(deck.main.size() < DECK_MAX_SIZE) {
 			deck.main.push_back(dataManager.GetCodePointer(code));
 		}
 	}
--- a/gframe/deck_manager.cpp
+++ b/gframe/deck_manager.cpp
@@ -246,7 +246,9 @@ bool DeckManager::LoadDeck(irr::gui::IGUIComboBox* cbCategory, irr::gui::IGUICom
 }
 FILE* DeckManager::OpenDeckFile(const wchar_t* file, const char* mode) {
 #ifdef WIN32
-	FILE* fp = _wfopen(file, (wchar_t*)mode);
+	wchar_t wmode[20]{};
+	BufferIO::CopyWStr(mode, wmode, sizeof(wmode) / sizeof(wchar_t));
+	FILE* fp = _wfopen(file, wmode);
 #else
 	char file2[256];
 	BufferIO::EncodeUTF8(file, file2);
--- a/gframe/deck_manager.h
+++ b/gframe/deck_manager.h
@@ -8,6 +8,10 @@
 #include <sstream>
 
 namespace ygo {
+	constexpr int DECK_MAX_SIZE = 60;
+	constexpr int DECK_MIN_SIZE = 40;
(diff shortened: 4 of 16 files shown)


salix5:
@edo9300 
I think edopro has the same problems too.
Can you take a look?
Thank you!

edo9300:
> @edo9300 I think edopro has the same problems too. Can you take a look? Thank you!

Thanks for the concern, in my case no static buffers are being used, so buffer overruns shouldn't happen in this scenario, _what would happen would be a bad alloc exception being raised and the program crashing if there won't be enough memory available for the vector_

salix5:
@edo9300 
https://github.com/edo9300/edopro/blob/master/gframe/netserver.h
It seems that `net_server_read` and `net_server_write` are still static arrays.

purerosefallen:
@salix5 In the past there are multiple cases that adding `std::` to `memcpy` would make it fail to build on Linux and Mac.

Could you please test them?

salix5:
> @salix5 In the past there are multiple cases that adding `std::` to `memcpy` would make it fail to build on Linux and Mac.
> 
> Could you please test them?
> 
> Also please do a favor to make a merge to mycard/ygopro:server.

Now `std::memcpy` is included in every file of gframe.
#include <cstring> in ocgcore/buffer.h
#include "../ocgcore/buffer.h" in gframe/bufferio.h
#include "bufferio.h" in gframe/config.h

gframe/deck_con.cpp
```diff
@@ -1757,7 +1757,7 @@ bool DeckBuilder::push_main(code_pointer pointer, int seq) {
 	if(pointer->second.type & (TYPE_FUSION | TYPE_SYNCHRO | TYPE_XYZ | TYPE_LINK))
 		return false;
 	auto& container = deckManager.current_deck.main;
-	int maxc = mainGame->is_siding ? 64 : 60;
+	int maxc = mainGame->is_siding ? 64 : DECK_MAX_SIZE;
```
purerosefallen, line 1760: Use dynamic way.

`int maxc = mainGame->is_siding ? DECK_MAX_SIZE + 4 : DECK_MAX_SIZE;`

purerosefallen (changes requested):
Please change it to dynamic way in deck_con.cpp

purerosefallen:
It seemed that the old client cannot connect to new server at all.

Related third-party clients should make some changes, including WindBot and YGOPro2. But we still had no idea on what to change.

DailyShana:
I think the main problem is this one PR do so many things, which make other maintainer hard to review.

The easy part like adding `std::` and use memcpy to avoid pointer cast can be split to another PR and be merged smoothly.

Checking length of deck data can be a another PR because it does not touch server things.
Checking lengh of STOC data should be another PR without changing structure of STOC
Although all above are about checking length of data, they are data of different facilities, split into different PR will make easier to review.

Changing or removing STOC structure will break compatibiliy, this part should be a single PR with fully discussion before merging.
By the way, I can't understand the reason above why to remove STOC_Chat

salix5:
> I think the main problem is this one PR do so many things, which make other maintainer hard to review.
> 
> The easy part like adding `std::` and use memcpy to avoid pointer cast can be split to another PR and be merged smoothly.
> 
> Checking length of deck data can be a another PR because it does not touch server things. Checking lengh of STOC data should be another PR without changing structure of STOC Although all above are about checking length of data, they are data of different facilities, split into different PR will make easier to review.
Maybe you can convince @purerosefallen  that we can open a new PR.

> Changing or removing STOC structure will break compatibiliy, this part should be a single PR with fully discussion before merging. By the way, I can't understand the reason above why to remove STOC_Chat
No, it does not.
Other clients still send the STOC_CHAT packet in the same way:
uint16_t  player_type, 2B
uint16_t msg, 1 ~ 256 code units

Casting to `struct STOC_Chat` is simply wrong.
The msg is not a fixed-size array `msg[256]`.
It never has been.
It is just a byte buffer which contains 1~256 `uint16_t` code units.

I removed it because no one really use that.

salix5:
#2555
remove the irrelevant commit
If you still think it is still too much, that it because ygopro has so many problems in the past ten years.

purerosefallen:
> #2555 remove the irrelevant commit If you still think it is still too much, that it because ygopro has so many problems in the past ten years.

I would continue to talk on this PR rather than #2551.
When we merge it we could squash it ignoring that tons of mid-commits.

In addition, some server implementations such as SRVPro, sends STOC_HS_PlayerEnter as 41 bytes without the tailing padding. Could we support those cases?

purerosefallen:
> I think the main problem is this one PR do so many things, which make other maintainer hard to review.
> 
> The easy part like adding `std::` and use memcpy to avoid pointer cast can be split to another PR and be merged smoothly.
> 
> Checking length of deck data can be a another PR because it does not touch server things. Checking lengh of STOC data should be another PR without changing structure of STOC Although all above are about checking length of data, they are data of different facilities, split into different PR will make easier to review.
> 
> Changing or removing STOC structure will break compatibiliy, this part should be a single PR with fully discussion before merging. By the way, I can't understand the reason above why to remove STOC_Chat

In case we do breaking changes we would change the version number.

In addition, if splitting a PR into multiple PRs, how do you make sure those PRs don't touch each other? It's even harder to review.

edo9300:
Fwiw, on my end i retrieve stoc structs this way 
```cpp
template<typename T>
static T getStruct(const void* data, size_t len) {
	T pkt{};
	memcpy(&pkt, data, std::min(sizeof(T), len));
	return pkt;
}
```
Allowing to perform short reads as well, leaving the rest of the struct zero initialized, in the case of STOC_Chat, it will support reading a chat message that doesn't send the whole buffer for example.

salix5:
> Fwiw, on my end i retrieve stoc structs this way
> 
> ```c++
> template<typename T>
> static T getStruct(const void* data, size_t len) {
> 	T pkt{};
> 	memcpy(&pkt, data, std::min(sizeof(T), len));
> 	return pkt;
> }
> ```
> 
> Allowing to perform short reads as well, leaving the rest of the struct zero initialized, in the case of STOC_Chat, it will support reading a chat message that doesn't send the whole buffer for example.

Thank you for your advice.

My opinion:
If the size of the `uint16_t` array is not fixed, we do not need a `struct` in the first place.
We can just keep the record of the max length of this array in `LEN_CHAT_MSG`.

```cpp
std::memcpy(player, pdata, LEN_CHAT_PLAYER*sizeof(uint16_t));
std::memcpy(msg, pdata, LEN_CHAT_MSG*sizeof(uint16_t));
```
Actually, it is much better than reading/writing `struct` to the network buffer directly because `struct` has padding.

edo9300:
Having a struct to define a network packet is way better than writing raw buffers, as you have a clear representation of what you're sending/receiving, in fact that's one of the main things that make the ygopro network infrastructure had to deal with. Regarding padding, that's really not an issue as it's clearly defined by the c/c++ standard, and sending 2 extra bytes won't cause any harm

salix5:
C++14 / N4140

5.3.3 Sizeof [expr.sizeof]
item 2
When applied to a class, the result is the number of bytes in an object of that class including any padding required for placing objects of that type in an array.


9.2 Class members [class.mem]
item 13
Nonstatic data members of a (non-union) class with the same access control (Clause 11) are allocated so that later members have higher addresses within a class object. 
The order of allocation of non-static data members with different access control is unspecified (Clause 11).
Implementation alignment requirements might cause two adjacent members not to be allocated immediately after each other; so might requirements for space for managing virtual functions (10.3) and virtual base classes (10.1).

It seems that the size of padding in a POD class is implementation-defined.
I did not find the explicit definition of size and location of padding in a POD class from the C++ standard.

@purerosefallen also mentions that 
```cpp
struct STOC_HS_PlayerEnter {
	unsigned short name[20];
	unsigned char pos;
	// byte padding[1]
};
static_assert(sizeof(STOC_HS_PlayerEnter) == 42, "size mismatch: STOC_HS_PlayerEnter");
```
The assertion may fail in some implementation.

salix5:
C++14 / N4140
3.9 Types [basic.types]
item 2
For any object (other than a base-class subobject) of trivially copyable type T, whether or not the object
holds a valid value of type T, the underlying bytes (1.7) making up the object can be copied into an array
of char or unsigned char.(42)
If the content of the array of char or unsigned char is copied back into theobject, the object shall subsequently hold its original value.

9 Classes [class]
item 6
A trivially copyable class is a class that:
(6.1) — has no non-trivial copy constructors (12.8),
(6.2) — has no non-trivial move constructors (12.8),
(6.3) — has no non-trivial copy assignment operators (13.5.3, 12.8),
(6.4) — has no non-trivial move assignment operators (13.5.3, 12.8), and
(6.5) — has a trivial destructor (12.4).

https://en.cppreference.com/w/cpp/types/is_trivially_copyable
template< class T >
struct is_trivially_copyable;

My conclusion:
If one really want to send a `struct` in a byte array, the minimal requirements are:
- It must be trivially copyable
checked by `static_assert(std::is_trivially_copyable<T>::value == true, "not trivially copyable")`

- The size must be a fixed value
checked by static_assert

- The padding must be the same
It seems that there are certain rules of data alignment in x86 system.
Other than that? Good luck.

- The endianness must be the same
(Almost) everyone uses little-endian system.
If not? Good luck.

Now we can safely write
```cpp
std::memcpy(&pkt, data, sizeof(T));
```

What happens if we write:
```cpp
std::memcpy(&pkt, data, sizeof(T)/2);
```

I do not know. Good luck.

purerosefallen:
Can we support STOC_HS_Player_Enter with 41 bytes for now? Then we may have some time to change every server implementation and deprecate that 41-byte message as time goes by.

salix5:
> Can we support STOC_HS_Player_Enter with 41 bytes for now? Then we may have some time to change every server implementation and deprecate that 41-byte message as time goes by.

done
It will check STOC_HS_PlayerEnter_size (41) for now.

purerosefallen:
Just a little more thought.

how about `#pragma pack(1)` on `STOC_HS_PlayerEnter`?

PR #2554, opened 2024-05-24
BufferIO: update CopyWStr

salix5:


purerosefallen:
What does it do?

salix5:
https://en.cppreference.com/w/cpp/string/byte/memcpy
https://en.cppreference.com/w/cpp/string/byte/strcpy
```cpp
void* memcpy( void* dest, const void* src, std::size_t count );
char* strcpy( char* dest, const char* src );
```

The source of many copy functions are pointers to `const T`.
It ensures that the source will not be modified in this function.

PR #2556, opened 2024-06-02
allow 9-digit code for lflist

purerosefallen:
This is very popular in custom card or rush duel communities. So we have to change it back.


salix5:
I did not notice that.
Thank you.

PR #2558, opened 2024-06-03
verify the source of BufferIO::CopyWStr

salix5:
https://en.cppreference.com/w/c/string/byte/strcpy

The source of BufferIO::CopyWStr must be null-terminated, just like `strcpy` in C.

@purerosefallen 
@mercury233

### DIFF
--- a/gframe/netserver.cpp
+++ b/gframe/netserver.cpp
@@ -250,7 +250,7 @@ void NetServer::HandleCTOSPacket(DuelPlayer* dp, unsigned char* data, int len) {
 		CTOS_PlayerInfo packet;
 		std::memcpy(&packet, pdata, sizeof packet);
 		auto pkt = &packet;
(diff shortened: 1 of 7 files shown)


mercury233:
shouldn't it be done in `BufferIO::CopyWStr` ?

gframe/netserver.cpp
```diff
@@ -249,7 +249,8 @@ void NetServer::HandleCTOSPacket(DuelPlayer* dp, unsigned char* data, int len) {
 			return;
 		CTOS_PlayerInfo packet;
 		std::memcpy(&packet, pdata, sizeof packet);
-		const auto* pkt = &packet;
+		auto pkt = &packet;
+		pkt->name[19] = 0;
```
mercury233, line 253: 19 -> sizeof -1

mercury233 (changes requested):

PR #2560, opened 2024-06-03
set max length of IGUIEditBox to LEN_CHAT_MSG - 1

salix5:
Avoid pasting very long text.

@mercury233 
@purerosefallen


salix5:
@mercury233 @purerosefallen 
Can we merge this one?

mercury233:
I think it should be set separately for each editbox

salix5:
The max length is unlimited now.
Adding a fixed max length is better than nothing.

PR #2568, opened 2024-07-20
clean config.h

salix5:
CGUITTFont.h
CGUIImageButton.h
mysignal.h
They are only used in `class game`.

myfilesystem.h
It is used in less than 10 files.
`#include "myfilesystem.h"` are moved to those files.

<time.h>
moved to `game.h`

<thread>
<mutex>
include when necessary

GL library
It is only used in `drawing.cpp`.
moved to `game.h`.

Add:
#include <string>
Some files might include <string> indirectly before.
The including path:
CGUITTFont.h
->irrUString.h
-> string

Rename guard macro:
https://gcc.gnu.org/onlinedocs/cpp/Once-Only-Headers.html
> In a user header file, the macro name should not begin with ‘_’. 
> In a system header file, it should begin with ‘__’ to avoid conflicts with user programs. 
> In any kind of header file, the macro name should contain the name of the file and some additional text, to avoid conflicts with other header files. 

Now the guard macro is YGOPRO_CONFIG_H following the GCC convention.
`#pragma once` is removed.

Test:
It can be built in VS 2022.
Maybe you can test it in GCC?

@mercury233 
@purerosefallen

### DIFF
--- a/gframe/client_card.cpp
+++ b/gframe/client_card.cpp
@@ -32,12 +32,12 @@ ClientCard::~ClientCard() {
 	}
 	overlayed.clear();
 }
-void ClientCard::SetCode(int code) {
-	if((location == LOCATION_HAND) && (this->code != (unsigned int)code)) {
-		this->code = code;
+void ClientCard::SetCode(int x) {
+	if((location == LOCATION_HAND) && (code != (unsigned int)x)) {
+		code = x;
 		mainGame->dField.MoveCard(this, 5);
 	} else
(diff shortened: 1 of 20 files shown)


purerosefallen:
Tests are needed on all **5** platforms: Windows, Linux, macOS (x86 and m1), server

purerosefallen:
Also please do the merge to https://github.com/mycard/ygopro/tree/server and submit a PR there.

PR #2569, opened 2024-07-20
use safe filename in Replay::SaveReplay

salix5:
Save the replay with filename `1*2*3`

Before:
The replay file is not created.

After:
The replay file is created with filename `1_2_3.yrp`.

@mercury233 
@purerosefallen


purerosefallen:
is it a security thing?

salix5:
I think no.

Current version:
If the input is an invalid filename, the replay file will not be stored.

salix5:
merge conflict 
temporarily closed

PR #2570, opened 2024-07-24
use UTF-8 locale

salix5:
https://learn.microsoft.com/en-us/cpp/porting/upgrade-your-code-to-the-universal-crt?view=msvc-170
The Microsoft C Runtime Library (CRT) was refactored in Visual Studio 2015. The Standard C Library, POSIX extensions and Microsoft-specific functions, macros, and global variables were moved into a new library, the Universal C Runtime Library (Universal CRT or UCRT). The compiler-specific components of the CRT were moved into a new vcruntime library.

The UCRT is now a Windows component, and ships as part of Windows 10 and later. The UCRT supports a stable ABI based on C calling conventions, and it conforms closely to the ISO C99 standard, with only a few exceptions.

Because the UCRT is now a Microsoft Windows operating system component, it's included as part of the operating system in Windows 10 and later. It's available through Windows Update for older operating systems, Windows Vista through Windows 8.1. A Redistributable version is available for Windows XP. As an operating system component, UCRT updates and servicing are managed by Windows Update independently of Visual Studio and Microsoft C++ compiler versions.

https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/setlocale-wsetlocale?view=msvc-170#utf-8-support
Starting in Windows 10 version 1803 (10.0.17134.0), the Universal C Runtime supports using a UTF-8 code page. The change means that char strings passed to C runtime functions can expect strings in the UTF-8 encoding. To enable UTF-8 mode, use ".UTF8" as the code page when using setlocale. For example, setlocale(LC_ALL, ".UTF8") uses the current default Windows ANSI code page (ACP) for the locale and UTF-8 for the code page.

----
Now we can use UTF8 locale on Windows, and the UCRT is available for older opearting systems too.
Now we can open all files with `fopen`.

@mercury233 
@purerosefallen

### DIFF
--- a/gframe/deck_manager.cpp
+++ b/gframe/deck_manager.cpp
@@ -247,15 +247,9 @@ bool DeckManager::LoadDeck(irr::gui::IGUIComboBox* cbCategory, irr::gui::IGUICom
 	return res;
 }
 FILE* DeckManager::OpenDeckFile(const wchar_t* file, const char* mode) {
-#ifdef WIN32
-	wchar_t wmode[20]{};
-	BufferIO::CopyWStr(mode, wmode, sizeof(wmode) / sizeof(wchar_t));
-	FILE* fp = _wfopen(file, wmode);
-#else
-	char file2[256];
+	char file2[256]{};
 	BufferIO::EncodeUTF8(file, file2);
 	FILE* fp = fopen(file2, mode);
-#endif
 	return fp;
 }
 IReadFile* DeckManager::OpenDeckReader(const wchar_t* file) {
--- a/gframe/gframe.cpp
+++ b/gframe/gframe.cpp
@@ -23,9 +23,7 @@ void ClickButton(irr::gui::IGUIElement* btn) {
 }
 
 int main(int argc, char* argv[]) {
-#ifndef _WIN32
-	setlocale(LC_CTYPE, "UTF-8");
-#endif
+	setlocale(LC_ALL, ".UTF-8");
 #ifdef __APPLE__
 	CFURLRef bundle_url = CFBundleCopyBundleURL(CFBundleGetMainBundle());
 	CFURLRef bundle_base_url = CFURLCreateCopyDeletingLastPathComponent(NULL, bundle_url);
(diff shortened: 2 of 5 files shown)


edo9300:
This change makes sense only if the project enforces using a statically linked ucrt while using a new enough version of the windows 10 sdk, otherwise it won't work and will actually break on systems with an older ucrt (if linked dynamically) or if built with an older sdk. It will also not work if using mingw to compile the application

salix5:
https://learn.microsoft.com/en-us/cpp/windows/universal-crt-deployment?view=msvc-170
Beginning in Visual Studio 2015, it's no longer the case. Visual Studio 2015 and later versions of Visual Studio all use one Universal CRT.
The Universal CRT (UCRT) is a Microsoft Windows operating system component. It's included as part of the operating system in Windows 10 or later, and Windows Server 2016 or later. The UCRT is available by using Windows Update on older operating systems that are still in extended support. Local deployment of the Universal CRT is supported, with some restrictions.

https://learn.microsoft.com/en-us/cpp/windows/universal-crt-deployment?view=msvc-170#central-deployment
The preferred method to centrally install the Universal CRT is to use Microsoft Windows Update. The Universal CRT is a Recommended update for all supported Microsoft Windows operating systems, so by default, most machines install it as part of the regular update process. The initial release of the Universal CRT was [KB2999226](https://support.microsoft.com/kb/2999226). A later update with various bug fixes was made in [KB3118401](https://support.microsoft.com/kb/3118401), and there have been more updates with further bug fixes and new features. For more recent updates, search [support.microsoft.com](https://support.microsoft.com/) for "Universal C Runtime" or "Universal CRT".

Not all Microsoft Windows computers regularly install updates by use of Windows Update, and some may not install all Recommended updates. To support the use of applications built by using the Visual Studio 2015 and later C++ toolsets on those machines, there are Universal CRT Redistributable files available for offline distribution. Those redistributable files may be downloaded from one of the KB links above. The Universal CRT Redistributable requires that the machine has been updated to the current service pack. So, for example, the redistributable for Windows 7 will only install onto Windows 7 SP1, not Windows 7 RTM.

----
Microsoft says that:
- UCRT is a Windows component, which is available through Windows Update for older operating systems.
- Even if some did not install Recommended updates for some reasons, there are Universal CRT Redistributable files available for offline distribution.

If these statements are true, the version of UCRT will not be a problem.

mercury233:
need test

https://github.com/Fluorohydride/ygopro/pull/2160

PR #2576, opened 2024-08-14
reorder config.h

mercury233:
`swprintf` is defined in `<string>`

### DIFF
--- a/gframe/config.h
+++ b/gframe/config.h
@@ -43,22 +43,22 @@
 #define mystrncasecmp strncasecmp
 #endif
 
+#include <string>
 template<size_t N, typename... TR>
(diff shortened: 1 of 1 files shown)


purerosefallen:
Can it fix the problem of building on Linux / macOS?

mercury233:
> Can it fix the problem of building on Linux / macOS?

maybe partly

purerosefallen:
> > Can it fix the problem of building on Linux / macOS?
> 
> maybe partly

Then we have to solve https://github.com/Fluorohydride/ygopro/issues/2578 to purpose another testing on mc/develop, otherwise it can't be released.

salix5:
@mercury233 
https://en.cppreference.com/w/cpp/io/c/fwprintf
std::swprintf
Defined in header <cwchar>

salix5:
#2581

PR #2577, opened 2024-08-14
add import/export deck code

mercury233:
image


purerosefallen:
Can we take a chance to make all distro's deck code the same format?

edo9300:
I would suggest to use the `ydke` url format, it's simpler than the other alternatives proposed (just straight up convert 3 arrays to base64), and it's been used for 4 years and most of the "western" ygopro tools/sites already have support for it

mercury233:
_and ygo233's_
https://ygo233.com/deck

purerosefallen:
> I would suggest to use the `ydke` url format, it's simpler than the other alternatives proposed (just straight up convert 3 arrays to base64), and it's been used for 4 years and most of the "western" ygopro tools/sites already have support for it

However, it's another url format in *eastern* ygopro things.



> _and ygo233's_ https://ygo233.com/deck

*then you do it*, none of us know what that is

edo9300:
> @edo9300 any docs or code to see?
> 
There's this JavaScript implementation of encoding/decoding a ydke url
https://github.com/ProjectIgnis/ydke.js/blob/master/src/index.ts

purerosefallen:
Hello?

purerosefallen:
https://code.moenext.com/mycard/ygopro-deck-encode

Let's consider those deck codes instead of tossing a whole YDK in

mercury233:
> https://code.moenext.com/mycard/ygopro-deck-encode
> 
> Let's consider those deck codes instead of tossing a whole YDK in

The raw YDK text is most common. I think it is not a good idea to support all of the existing ydk shortener, nor inventing a "prefect" one yourself and let everyone use that.

purerosefallen:
> > https://code.moenext.com/mycard/ygopro-deck-encode
> > Let's consider those deck codes instead of tossing a whole YDK in
> 
> The raw YDK text is most common. I think it is not a good idea to support all of the existing ydk shortener, nor inventing a "prefect" one yourself and let everyone use that.

This includes the following:
- A new one
- YDKe (the Ignis one)
- YGOMobile one
- `CTOS_UPDATE_DECK` format

purerosefallen:
> > https://code.moenext.com/mycard/ygopro-deck-encode
> > Let's consider those deck codes instead of tossing a whole YDK in
> 
> The raw YDK text is most common. I think it is not a good idea to support all of the existing ydk shortener, nor inventing a "prefect" one yourself and let everyone use that.

Or support one of them?

PR #2582, opened 2024-08-18
 add wrapper function myfopen 

salix5:
inline FILE* myfopen(const char* filename, const char* mode)
The wrapper function of `fopen`.

Replay::SaveReplay
Replay::OpenReplay
Now they use local `FILE*` variable.
`Replay::fp` is only used in replay recording.

@mercury233 
@purerosefallen


salix5:
There are several problems in UTF8 functions.
Waiting for #2585

PR #2585, opened 2024-08-31
 fix EncodeUTF8String, DecodeUTF8String

salix5:
UTF-16
High surrogate: 0xD800 ~ 0xDBFF
Low surrogate: 0xDC00 ~ 0xDFFF
Surrogate pair: 1 High + 1 Low

EncodeUTF8String 
Windows: sizeof(wchar_t)=2, wchar_t[] is treated as UTF-16
Other: sizeof(wchar_t)=4, wchar_t[] is treated as UTF-32

On Windows
- If the last code unit in wchar_t[] is a lone surrogate, it will read out-of bounds.
It will try to read a[10] in this example:
```cpp
wchar_t a[10]={0xd800,0xd800,0xd800,0xd800,0xd800,0xd800,0xd800,0xd800,0xd800,0}; 
char dest[256]{};
EncodeUTF8String(a, dest, sizeof dest);
```

- It does not check surrogate pairs.
2 lone surrogate: 0xd800,0xd800
Low surrogate first: 0xdc00, 0xd800
It will accept these sequences, which are ill-formed UTF-16 sequences.

DecodeUTF8String
- If the last byte in char[] is a leading byte, it will read out-of bounds.
Example:
```cpp
char a[3] = {0xE0, 0xA4, 0};
wchar_t dest[256]{};
DecodeUTF8String(a, dest, 256);
```

- If the decoded Unicode code point is in Surrogate Area, it will write a lone surrogate.
Example:
UTF8: 0xED 0xA0 0x80
Unicode code point: High Surrogates (U+D800)
It will write a lone surrogate, but it shouldn't.

Solution:
```cpp
template<typename T>
bool CheckUTF8Byte(const T* str, int len)
```
Check if there are `len` continuation byte in the UTF-8 string.

```cpp
bool IsUnicodeChar(unsigned int c)
```
Check if the code point is a Unicode character.

@mercury233 
@purerosefallen 

@edo9300 
I believe that the UTF8 functions in edopro have the same problems.

Reference:
https://en.wikipedia.org/wiki/UTF-8
https://en.wikipedia.org/wiki/UTF-16
https://www.compart.com/en/unicode

### DIFF
--- a/gframe/bufferio.h
+++ b/gframe/bufferio.h
@@ -152,9 +152,37 @@ public:
 		const char* p = src;
 		wchar_t* wp = wstr;
 		while(*p != 0) {
-			const unsigned cur = *p & 0xffU;
+			unsigned cur = 0;
 			int codepoint_size = 0;
-			if ((cur & 0xf8) == 0xf0) {
+			if ((p[0] & 0x80U) == 0) {
+				cur = p[0] & 0xffU;
+				p++;
+			}
+			else if ((p[0] & 0xe0U) == 0xc0U) {
+				if (!CheckStringSize(p, 2))
+					break;
+				cur = ((p[0] & 0x1fU) << 6) | (p[1] & 0x3fU);
+				p += 2;
+			}
+			else if ((p[0] & 0xf0U) == 0xe0U) {
+				if (!CheckStringSize(p, 3))
+					break;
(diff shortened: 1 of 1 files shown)


edo9300:
> @edo9300 I believe that the UTF8 functions in edopro have the same problems.
> 

Indeed on my end as well the checks are only done on the output, and not on the input, thanks for finding it

salix5:
update:
Now it will check every continuation byte and overlong encoding.

PR #2593, opened 2024-10-03
 remove DataManager::GetNumString

salix5:
Using the same global buffer is bad, and it is totally unnecessary.
It is usually called by
```cpp
void DrawShadowText(irr::gui::CGUITTFont* font, const core::stringw& text, ...
```

It can be replaced by the wrapper function
```cpp
void DrawShadowText(irr::gui::CGUITTFont* font, int number, ...
```

`DrawShadowText(font, dataManager.GetNumString(x), ...`
It is replaced by
`DrawShadowText(font, x, ...`

other
Generate a local buffer by `myswprintf`.

@mercury233 
@purerosefallen

### DIFF
--- a/gframe/data_manager.cpp
+++ b/gframe/data_manager.cpp
@@ -117,8 +117,6 @@ bool DataManager::LoadStrings(const char* file) {
 		ReadStringConfLine(linebuf);
 	}
 	fclose(fp);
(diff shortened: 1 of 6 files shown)


mercury233:
need test for performance

PR #2595, opened 2024-10-05
fix DataManager::Format*

salix5:
Storing buffer in global array will cause race condition.
If the function is called from 2 different places, the result might be wrong.

@mercury233 
@purerosefallen


mercury233:
need test for performance

salix5:
wait #2597

PR #2632, opened 2024-12-12
 set info to default value before creating duel_mode 

salix5:
# Case 1
```cpp
CTOS_CreateGame packet;
packet.info.mode = 0xff;
```
In NetServer::HandleCTOSPacket
duel_mode will be nullptr
dereference nullptr in 
```cpp
duel_mode->host_info = pkt->info;
```

# Case 2
If deckManager._lfList is empty
dereference nullptr in 
```cpp
pkt->info.lflist = deckManager._lfList[0].hash;
```

@mercury233 
@purerosefallen 
@Wind2009-Louse
@fallenstardust


salix5:
merge because this is the smallest one of the security issues

這已經是已知的問題裡面最小的
因此直接合併

PR #2642, opened 2025-01-04
 fix ScriptReader 

salix5:
scriptBuffer
increased to 1MB

### DIFF
--- a/gframe/data_manager.cpp
+++ b/gframe/data_manager.cpp
@@ -6,7 +6,7 @@
 namespace ygo {
 
 const wchar_t* DataManager::unknown_string = L"???";
-unsigned char DataManager::scriptBuffer[0x20000];
+unsigned char DataManager::scriptBuffer[0x100000];
 IFileSystem* DataManager::FileSystem;
 DataManager dataManager;
 
--- a/gframe/data_manager.cpp
+++ b/gframe/data_manager.cpp
@@ -424,14 +424,11 @@ unsigned char* DataManager::ScriptReader(const char* script_name, int* slen) {
 #endif
 	if (!reader)
 		return nullptr;
-	size_t size = reader->getSize();
-	if (size > sizeof scriptBuffer) {
-		reader->drop();
-		return nullptr;
-	}
-	reader->read(scriptBuffer, size);
+	int size = reader->read(scriptBuffer, sizeof scriptBuffer);
 	reader->drop();
-	*slen = (int)size;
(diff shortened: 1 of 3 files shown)


gframe/data_manager.cpp
```diff
@@ -424,14 +424,11 @@ unsigned char* DataManager::ScriptReader(const char* script_name, int* slen) {
...
-	size_t size = reader->getSize();
-	if (size > sizeof scriptBuffer) {
-		reader->drop();
-		return nullptr;
-	}
-	reader->read(scriptBuffer, size);
+	int size = reader->read(scriptBuffer, sizeof scriptBuffer);
 	reader->drop();
-	*slen = (int)size;
+	if (size >= (int)sizeof scriptBuffer)
```
mercury233, line 429: Is this possible?

salix5: Yes.
read:
1.lua,  size 1145141919810 bytes
size = sizeof scriptBuffer

mercury233: `reader->read` will return count of read bytes, which will be less than the param `sizeof scriptBuffer`, so I think this line won't make sense

salix5: https://irrlicht.sourceforge.io/docu/classirr_1_1io_1_1_i_read_file.html#ab51878d36bc9dd3964b664055fbeb13f
sizeToRead
Amount of bytes to read from the file. 

Returns:
How many bytes were read. 

https://en.cppreference.com/w/cpp/io/c/fread
count
the number of the objects to be read

Return value
Number of objects read successfully, which may be less than count if an error or end-of-file condition occurs. 

If the size of file is larger than `scriptBuffer`, both of them will return `sizeof scriptBuffer`.
Please check the document first.

mercury233: Do you mean `size` will never be greater than `sizeof scriptBuffer`? So why we need this check?

salix5: sizeof scriptBuffer
目前是1048576

檔案大小X
if X < 1048576
fread返回 X

if X >= 1048576
fread返回1048576

irrlicht的read
是fread的推廣版本

可能返回值：
1048575 (完全讀取)

1048576 (讀取到達buffer上限)
可能被截斷故拒絕

我不太明白問題在哪...

mercury233: 之前的版本有什么问题

salix5: 你不如先說你覺得還有什麼問題

PR #2643, opened 2025-01-06
client MSG_SELECT_SUM: cancellable & better display

purerosefallen:
image

It displays the value of current sum, and it can be cancelled, just like how Group.SelectSubGroup does.

The following lines of `strings.conf` are no longer needed and can be removed:

```
!system 209 当前所选的卡已满足选择条件
!system 210 是否要继续选择？
```


Wind2009-Louse:
Test passed

PR #2651, opened 2025-01-14
use <cwchar> function

salix5:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/snprintf-snprintf-snprintf-l-snwprintf-snwprintf-l?view=msvc-170
Beginning with the UCRT in Visual Studio 2015 and Windows 10, snprintf is no longer identical to _snprintf.
The snprintf behavior is now C99 standard conformant.

Now we can use <cwchar> functions.

@mercury233 
@purerosefallen 
@fallenstardust

@Wind2009-Louse
可否幫忙測試在其他平台編譯的情形
謝謝

### DIFF
--- a/gframe/config.h
+++ b/gframe/config.h
@@ -43,12 +43,6 @@
 #define mystrncasecmp strncasecmp
 #endif
 
-#include <wchar.h>
-template<size_t N, typename... TR>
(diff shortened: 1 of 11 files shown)


Wind2009-Louse:
Successfully compiled on Windows, Linux and Macos

PR #2667, opened 2025-02-09
remove irrlicht/irrlicht.patch

salix5:
@mercury233 
irrlicht 1.8.5
細節的問題很多
以後還有可能發現其他的問題
可能已經遠超過可以用單一檔案patch的範圍

我個人建議
編譯說明改成直接使用修正後的irrlicht
irrlicht.patch
從此不再維護


purerosefallen:
Why not adding it to this repo?

purerosefallen:
Btw, can it compile with the original irrlicht? no matter if ygopro runs well

salix5:
> Btw, can it compile with the original irrlicht? no matter if ygopro runs well

`std::snprintf`
`std::swprintf`
These expressions cannot be used in the original irrlicht 1.8.5.
So https://github.com/mercury233/irrlicht/pull/4 is necessary.

And in case you might ask:
Using `std::snprintf` is correct and recommended by C++ standard.
If you cannot compile while using the correct expression, it is the problem of the ones who add these macro.
irrlicht should have removed these sections 10 years ago, but they didn't.

PR #2669, opened 2025-02-12
use const reference in is_setcodes

salix5:
@mercury233


purerosefallen:
What is the improvement?

PR #2684, opened 2025-02-18
use <cstdio> functions

salix5:


purerosefallen:
how is it different?

Note: how did you do the replacement? There are other forks such as server.

salix5:
> how is it different?

<stdio.h> is deprecated long ago.
The standard encourages using the functions in namespace std.

> 
> Note: how did you do the replacement? There are other forks such as server.

fread, fwrite 
fopen, fclose, fgets 
fflush
snprintf, sscanf
fprintf

replaced by
std::fread, std::fwrite
and so on

yybbwc:
> <stdio.h> is deprecated long ago.
> The standard encourages using the functions in namespace std.

c++23不是取消弃用了吗？

yybbwc:
https://cppreference.cn/w/cpp/header

> 注意：xxx.h 头文件在 C++98 中已弃用，并在 C++23 中重新启用。对于纯粹的 C++ 代码，这些头文件不建议使用，但不会被移除。

salix5:
For some of the C standard library headers of the form xxx.h, the C++ standard library both includes an identically-named header and another header of the form cxxx (all meaningful cxxx headers are listed above). The intended use of headers of form xxx.h is for interoperability only. It is possible that C++ source files need to include one of these headers in order to be valid ISO C. Source files that are not intended to also be valid ISO C should not use any of the C headers.

PR #2693, opened 2025-02-22
fix for latest freetype

mercury233:
fix: freetype.lib(ftinit.obj) : error LNK2001: unresolved external symbol _ft_svg_renderer_class 
Or maybe we can use some switch to disable the svg feature of freetypee

### DIFF
--- a/premake/freetype/premake5.lua
+++ b/premake/freetype/premake5.lua
@@ -43,6 +43,10 @@ project "freetype"
             "src/type42/type42.c",
             "src/winfonts/winfnt.c" }
 
+    if os.isfile("src/svg/svg.c") then
+        files { "src/svg/svg.c" }
+    end
+
     filter "system:windows"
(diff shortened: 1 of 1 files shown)


purerosefallen:
Is it compatitable with the prev version of freetype?

mercury233:
> Is it compatitable with the prev version of freetype?

Copilot:
> The highlighted line in the provided diff adds a new conditional check to include the file `src/svg/svg.c` only if it exists:
> ```lua
> if os.isfile("src/svg/svg.c") then
>     files { "src/svg/svg.c" }
> end
> ```
> This change is compatible with the previous version of FreeType because it does not alter existing functionality. Instead, it conditionally includes an additional file without affecting the build if the file does not exist. This ensures that the change is backward-compatible and will not cause any issues with previous versions that do not have the `src/svg/svg.c` file.

yybbwc:
好好好，你们终于用上新版freetype了

PR #2703, opened 2025-02-27
加个保护

yybbwc:


mercury233:
please explain what you want and what this will do

yybbwc:
> please explain what you want and what this will do

自动把未初始化的变量初始化为零。

yybbwc:
https://source.android.google.cn/docs/security/test/memory-safety/zero-initialized-memory?hl=zh-cn

edo9300:
The variables should just bee explicitly initialized rather that relying on compiler specific behavior

yybbwc:
> The variables should just bee explicitly initialized rather that relying on compiler specific behavior

It merely serves as a fallback protection and does not affect the initialization of explicitly declared variables.

edo9300:
What should be added is `-Wuninitialized / -Wmaybe-uninitialized` and actually handle the warning generated

yybbwc:
> What should be added is `-Wuninitialized / -Wmaybe-uninitialized` and actually handle the warning generated

Oh, right, you're correct. I usually have too many warnings when writing code, so I'm accustomed to turning off the warnings for compilation... I'm not very familiar with the parameters for warnings.

purerosefallen:
Have you tested on Linux and Mac

yybbwc:
> Have you tested on Linux and Mac

没有。

purerosefallen:
Could lead to performance issues.

Sometimes

```cpp
uint8_t some_buf[10];
query_field_card(some_buf);
```

it doesn't matter if the vars iniitalized or not.

I believe the ygopro code handles them correctly, and not needed to do them in compile options.

PR #2707, opened 2025-03-01
fix open ydk, yrp file

salix5:
ydk file:
-ehero.ydk

`YGOPro.exe -ehero.ydk`

Before:
treated as `-e${cdb_name}`
the file is not opened

After:
opening ydk/yrp file has highest priority
the file is opened

@mercury233 
@purerosefallen 
@Wind2009-Louse
@fallenstardust


mercury233:
_had better rework those part, `ClickButton` is not good_

salix5:
> _had better rework those part, `ClickButton` is not good_

I will try when I have time.

PR #2711, opened 2025-03-02
fix

yybbwc:
修复由小海豚引入的bug。
#2691 
原有实现导致无法加载带中文名的残局


mercury233:
I think the problem only happen on windows

yybbwc:
> I think this problem only happen on windows

当然是发生在windows啊。因为windows不用utf8。
哦，对了，decode时会发生类型转换，导致在非windows平台出错。

yybbwc:
#2654
引入的

purerosefallen:
You have to fix `ReadScriptFromFile` instead of changing this line.

mercury233:
fixed in https://github.com/Fluorohydride/ygopro/pull/2714

PR #2718, opened 2025-03-08
move to C++17

salix5:
https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170
Most C++17 features are supported in VS2017~2019, which is at least 6 years ago.
Maybe it is time to move to C++17.

(update the conclusion in #2603)

some C++17 feautures we may need:
- new helper functions
- constexpr if 
- terse static assert
- selection statements with initializer

related:
https://github.com/mercury233/irrlicht/pull/7

@mercury233 
@purerosefallen 
@Wind2009-Louse
@fallenstardust

### DIFF
--- a/gframe/CGUIImageButton.cpp
+++ b/gframe/CGUIImageButton.cpp
@@ -313,8 +313,8 @@ void CGUIImageButton::draw() {
 		return;
 	IGUISkin* skin = Environment->getSkin();
 	video::IVideoDriver* driver = Environment->getVideoDriver();
-	core::vector2di center = AbsoluteRect.getCenter();
-	core::vector2di pos = center;
+	irr::core::vector2di center = AbsoluteRect.getCenter();
+	irr::core::vector2di pos = center;
 	pos.X -= (s32)(ImageRect.getWidth() * imageScale.X * 0.5f);
 	pos.Y -= (s32)(ImageRect.getHeight() * imageScale.Y * 0.5f);
 	if(Pressed) {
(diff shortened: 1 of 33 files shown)


purerosefallen:
Which version of GCC, or Linux distro, supports C++ 17? We have to consider server OS things.

purerosefallen:
Any required changes on premake script?

salix5:
https://gcc.gnu.org/projects/cxx-status.html#cxx17
GCC has almost full support for the 2017 revision of the C++ standard. The status of C++17 library features is described in [the library documentation](https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2017).

C++17 mode is the default since GCC 11




> Any required changes on premake script?
No

edo9300:
For premake, you should use the cppdialect option to pass the language version, and will take care of setting the compiler flags (there's an option for rtti as well)

salix5:
> For premake, you should use the cppdialect option to pass the language version, and will take care of setting the compiler flags (there's an option for rtti as well)

Thank you!
I don't know much about the flag "-fno-rtti", so I keep it for now.


@purerosefallen 
The premake scripts are updated.

purerosefallen:
does it work now?

salix5:
yes

yybbwc:
怎么还没合并……

mercury233:
I think what we should do is fix the code to be compatible with the new C++ standard, rather than breaking compatibility with the old standard just to use some new features for non-essential functionality.

salix5:
> rather than breaking compatibility with the old standard just to use some new features for non-essential functionality.

like what?
can you give an example?

mercury233:
> like what? can you give an example?

constexpr if

I guess it will be optimized by compiler already

salix5:
For the record, using new features in C++17 is not "breaking compatibility".
I guess people call it progress.

PR #2731, opened 2025-03-16
add rtti option

salix5:
@mercury233
@purerosefallen
@Wind2009-Louse
@fallenstardust

### DIFF
--- a/gframe/premake5.lua
+++ b/gframe/premake5.lua
@@ -4,6 +4,7 @@ include "spmemvfs/."
 project "YGOPro"
     kind "WindowedApp"
     cppdialect "C++14"
+    rtti "Off"
 
     files { "*.cpp", "*.h" }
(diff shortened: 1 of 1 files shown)


purerosefallen:
What's that?

salix5:
https://premake.github.io/docs/rtti/

https://gcc.gnu.org/onlinedocs/gcc-4.4.7/gcc/C_002b_002b-Dialect-Options.html
-fno-rtti
Disable generation of information about every class with virtual functions for use by the C++ runtime type identification features (`dynamic_cast' and `typeid').
If you don't use those parts of the language, you can save some space by using this flag.
Note that exception handling uses the same information, but it will generate it as needed. 
The `dynamic_cast' operator can still be used for casts that do not require runtime type information, i.e. casts to void * or to unambiguous base classes.

PR #2733, opened 2025-03-17
Fix premake5 on beta1

purerosefallen:


mercury233 (approved):
> `linktimeoptimization "value"`
Availability
Premake 5.0-beta4 and later
https://premake.github.io/docs/linktimeoptimization/

Premake 5.0-beta4 uses new glibc, so some systems such as Ubuntu 20.04 can't run.

Maybe we should support earlier premake.

purerosefallen:
> > `linktimeoptimization "value"`
> > Availability
> > Premake 5.0-beta4 and later
> > https://premake.github.io/docs/linktimeoptimization/
> 
> Premake 5.0-beta4 uses new glibc, so some systems such as Ubuntu 20.04 can't run.
> 
> Maybe we should support earlier premake.

beta5 is even worse. Even my Ubuntu 22.04 laptop can't run.

In addition, mycard ygopro linux build uses Debian 10, for an early-as-possible glibc support. It should support that situation.

purerosefallen:
Btw what is the current minimum version requirement of premake5?

PR #2738, opened 2025-03-22
move files & add copy.sh

salix5:
cmake/CMakeLists.txt
outdated
move to cmake/ for backup

resource/gframe/ygopro.ico
resource/gframe/ygopro.rc
irrelevant to premake5
move to another directory

copy.sh
copy premake5 files

@mercury233 
@purerosefallen 
@Wind2009-Louse
@fallenstardust


copy.sh
```diff
@@ -0,0 +1 @@
+cp -r premake/* .
```
purerosefallen, line 1: add `#!/bin/sh` on the 1st line, and make sure it's 0755 file mode.

.github/workflows/build.yml
```diff
@@ -228,6 +229,7 @@ jobs:
     - name: Copy premake files
       run: |
         cp -r premake/* .
+        cp -r resource/* .
```
purerosefallen, line 232: Don't forget macOS section

purerosefallen (changes requested):

purerosefallen:
I suggest to use `#!/bin/sh` on super-simple shell scripts because some distros like Alpine does not have bash by default.

.github/workflows/build.yml
mercury233, line 232: https://stackoverflow.com/questions/1997172/is-there-a-linux-equivalent-of-windows-resource-files

I don't think Linux need resource files

PR #2741, opened 2025-03-23
move cdialect

salix5:


salix5:
https://learn.microsoft.com/en-us/lifecycle/products/windows-xp
Windows XP
Extended End Date: Apr 8, 2014

It has been more than 10 years since the end of support of Windows XP.
Maybe it is time to leave.

salix5:
@mercury233 
@purerosefallen 
@Wind2009-Louse
@fallenstardust

purerosefallen:
There are still Windows XP users.

salix5:
We cannot wait forever.
If 10 years after the end of lifecycle is not long enough, then how long should we wait?

purerosefallen:
> We cannot wait forever. If 10 years after the end of lifecycle is not long enough, then how long should we wait?

Until there are hardly Windows XP users in user base.

mercury233:
Actually, the “official” release on ys168.com dropped WinXP support years ago, and no users have posted any questions about it there.

However, I don't think we should drop support for old systems just because they are old. The real reasons for discontinuing support are difficulties in maintenance, adding new features, replacing insecure APIs, upgrading dependencies, and other related issues.

salix5:
/permissive-
It cannot be used under `toolset "v140_xp"`, because Windows SDK 7.1A itself does not follow the standard.
And we are going to add more 
```lua
if not WINXP_SUPPORT then
```
sooner or later.

PR #2742, opened 2025-03-23
use copy.sh in actions build

purerosefallen:


purerosefallen:
> /home/runner/work/_temp/fd0e8c0d-e8c9-48eb-a446-a2598fc42b60.sh: line 1: ./copy.sh: Permission denied

@salix5 remember to change copy.sh to 0755

salix5:
It is not designed for Github action.
It is for regenerate premake5 files locally.
Why would you want to load a script with 1 line `cp` in Github action?

purerosefallen:
> It is not designed for Github action. It is for regenerate premake5 files locally. Why would you want to load a script with 1 line `cp` in Github action?

1. GitHub actions should check if everything works, including this `copy.sh` file. This is in fact a part of the coverage test.
2. If the action pipeline fails, it means the `copy..sh` script is broken with insufficient file mode, expected to be 0755. I'm creating this PR to check if this file actually works.

purerosefallen:
Plus, why not also copying the `resources` dir out?

purerosefallen:
> It is for regenerate premake5 files locally.

Why don't you make it a `build.sh`, copying everything out, then premake, and build?

```shell
#!/bin/sh

# Environment variable directory list
COPY_OUT_DIRS=(
  "premake"
  "resource"
)

# Loop through environment variables and copy the files
for dir in "${COPY_OUT_DIRS[@]}"; do
  if [ -d "$dir" ]; then
    cp -r "$dir"/* .
  else
    echo "Directory $dir does not exist, skipping..."
  fi
done

# Possible paths for PREMAKE5_BIN
PREMAKE5_BIN_POSSIBLE_PATHS=(
  "./premake5"
  "/usr/local/bin/premake5"
  "/usr/bin/premake5"
)

# Try to find the premake5 binary in the possible paths
PREMAKE5_BIN=""
for path in "${PREMAKE5_BIN_POSSIBLE_PATHS[@]}"; do
  if [ -x "$path" ]; then
    PREMAKE5_BIN="$path"
    break
  fi
done

# If PREMAKE5_BIN is still empty, echo an error and exit
if [ -z "$PREMAKE5_BIN" ]; then
  echo "Error: Could not find premake5 binary!"
  exit 1
fi

# Run the premake5 command
"$PREMAKE5_BIN" gmake "$@"

# Build the project
cd build
make config=release -j$(nproc)
cd ..
```

salix5:
1. No, it is not. 
It is just a simple script saving the time of ctrl+A, ctrl+C, ctrl+V.
If you cannot accept it, we can delete it and put it in `.gitignore`.

2. If the action pipeline fails, it means that we should write `cp` command directly.
The file permission should be determined by anyone who download this file.

3. Because many branches have their own resource file.
Resource files should not be overwritten when regenerating premake5 files.

salix5:
It can be run in Git bash of 
git version 2.48.1.windows.1

So no, it is not broken.
People who want to execute it should find a way to set the permission by themselves.

PR #2745, opened 2025-03-23
chmod=+x

salix5:


purerosefallen:
https://github.com/Fluorohydride/ygopro/pull/2742

Could you add it in actions to make sure it works?

salix5:
I agree that I should commit a script file with `chmod=+x`, and that is what I did.
But I don't think the Github action needs to be modified.
I will try to investigate it when I have time.

purerosefallen:
> I agree that I should commit a script file with `chmod=+x`, and that is what I did. But I don't think the Github action needs to be modified. I will try to investigate it when I have time.

That PR was because I wish to check if this script works, as GitHub does not show file mode in PRs.

PR #2747, opened 2025-03-24
Delete copy.sh

mercury233:
> If you cannot accept it, we can delete it and put it in .gitignore.
https://github.com/Fluorohydride/ygopro/pull/2742

@salix5 I don't think we should add this file.
Evenmore, you shouldn't add this to .gitignore, because this file is only used by you, and git can ignore file locally (.git/info/exclude)

### DIFF
--- a/copy.sh
+++ b/copy.sh
@@ -1,2 +1,0 @@
-#!/bin/sh
-cp -r premake/* .


salix5:
> > If you cannot accept it, we can delete it and put it in .gitignore.
> > #2742
> 
> @salix5 I don't think we should add this file. Evenmore, you shouldn't add this to .gitignore, because this file is only used by you, and git can ignore file locally (.git/info/exclude)

It is not only used by me.

The build instructions in wiki said that :
Copy all folders in the premake folder to the project folder, which contains the premake description files of each dependent library, which can be modified and improved by yourself

Everyone has to copy the premake folders when they create the project file.
Are they supposed to copy the folders manually?
It is 2025 now.

mercury233:
I think "copy" is very common when starting a new project.

Professional users will check the script's content before running it, and when they see that it contains only a single line, they may find it annoying. On the other hand, non-professional users, especially Windows users, may find running an sh script itself more difficult than simply copying some files.

salix5:
It can 

> I think "copy" is very common when starting a new project.
> 
> Professional users will check the script's content before running it, and when they see that it contains only a single line, they may find it annoying. On the other hand, non-professional users, especially Windows users, may find running an sh script itself more difficult than simply copying some files.

If one did run the script, why would he find it annoying because it has only 1 line?
Completing the job with minimal lines is an advantage, not disadvantages.

For window users, the script can be run by double clicking if they have installed git.
How could double clicking "more difficult" than:
going into premake
select all 
copy
going back to root directory 
paste

purerosefallen:
```
#!/bin/bash
set -x 
set -o errexit

export MESSAGE_DEBUG=1

export EVENT_INCLUDE_DIR=$PWD/libevent-stable/include
export EVENT_LIB_DIR=$PWD/libevent-stable/lib

# Environment variable directory list
COPY_OUT_DIRS=(
  "premake"
  "resource"
)

# Loop through environment variables and copy the files
for dir in "${COPY_OUT_DIRS[@]}"; do
  if [ -d "$dir" ]; then
    cp -r "$dir"/* .
  else
    echo "Directory $dir does not exist, skipping..."
  fi
done

# Possible paths for PREMAKE5_BIN
PREMAKE5_BIN_POSSIBLE_PATHS=(
  "./premake5"
  "/usr/local/bin/premake5"
  "/usr/bin/premake5"
)

# Try to find the premake5 binary in the possible paths
PREMAKE5_BIN=""
for path in "${PREMAKE5_BIN_POSSIBLE_PATHS[@]}"; do
  if [ -x "$path" ]; then
    PREMAKE5_BIN="$path"
    break
  fi
done

# If PREMAKE5_BIN is still empty, echo an error and exit
if [ -z "$PREMAKE5_BIN" ]; then
  echo "Error: Could not find premake5 binary!"
  exit 1
fi

"$PREMAKE5_BIN" gmake --build-freetype --build-sqlite --build-irrlicht --no-lua-safe "$@"

cd build
make config=release -j$(nproc)
cd ..

strip ygopro
```

Just to tell you, this is my `build.sh`

purerosefallen:
> I think "copy" is very common when starting a new project.
> 
> Professional users will check the script's content before running it, and when they see that it contains only a single line, they may find it annoying. On the other hand, non-professional users, especially Windows users, may find running an sh script itself more difficult than simply copying some files.

~~copy.bat~~

mercury233:
> If one did run the script, why would he find it annoying because it has only 1 line?
"You should tell me to copy some files, but you let me run a script"

> For window users, the script can be run by double clicking if they have installed git.
Many editors will assoc .sh files on Windows.

yybbwc:
顺便说一句，msys2也能运行sh文件

mercury233:
https://github.com/Fluorohydride/ygopro/commit/a63484208e5c2ac31080717ee871de9633aae358

PR #2755, opened 2025-04-05
update irrlicht premake

mercury233:
- Remove defines.lua
- Add libpng SSE2 optimization
- Add NO_IRR_COMPILE_WITH_ZIP_ENCRYPTION_, remove aesGladman things
- Don't need to link any lib under DXSDK_DIR


salix5 (approved):
行吧

PR #2756, opened 2025-04-14
textfont and numfont support spaces

mercury233:


salix5 (changes requested):
gameConf.textfontsize = std::strtol(last_space + 1, nullptr, 10);

My suggestion:
```cpp
int size = std::strtol(last_space + 1, nullptr, 10);
if (size <= 0)
	continue;
gameConf.textfontsize = size;
```
(skip if strtol fails)

mercury233:
`gameConf.textfontsize` is unsigned char, so "overflow" is somehow safe.
https://stackoverflow.com/questions/18195715

The user will only see weird sized fonts if this value is wrong (except 0, will fix in another PR)

PR #2761, opened 2025-04-19
add support for load-once.conf

mercury233:
我们可以将 `system.conf` 从版本控制中移除，改为直接写入源码中。  
目前，大多数发行版在打包完整安装包时都会包含 `system.conf` 文件，这导致用户在重新安装完整版时会丢失原有的自定义配置。  
如果某个发行版需要使用不同于源码的默认配置，可以改为通过 `load-once.conf` 文件来实现。程序会在启动时读取该文件中的配置内容，并自动删除它。  
此外，像 Windbot 这样的扩展也可以利用 `load-once.conf` 文件来更新其所需的配置。

---

We can remove `system.conf` from version control and instead write its contents directly into the source code.  
Currently, most distributions will include the `system.conf` file in the full installation package, which causes users to lose their custom configurations when reinstalling the full version.  
If a distribution requires default settings that differ from those in the source code, it can use a `load-once.conf` file instead. The program will read this file at startup and automatically delete it afterward.  
Additionally, extensions like Windbot can also use `load-once.conf` to apply the configurations they need.


purerosefallen:
Why delete?
The philosophy should be further discussed.

purerosefallen:
In the past, my distro uses `system_user.conf` as user-saved configuration settings, just something like SRVPro.

yybbwc:
懂了，用于更新

purerosefallen:
> Additionally, extensions like Windbot can also use load-once.conf to apply the configurations they need.

How to? Plz describe on this thought.

purerosefallen:
We may need two files, `system-default.conf` and `load-once.conf`

`system-default.conf` is for distro default config, and `load-once.conf` is for apply and then delete.

mercury233:
> > Additionally, extensions like Windbot can also use load-once.conf to apply the configurations they need.
> 
> How to? Plz describe on this thought.

```
enable_bot_mode = 1
bot_deck_path = ./windbot/Decks
```

A user’s downloaded version of YGOPro might not originally have the local bot functionality. When they decide to install Windbot, they need to modify these two lines in the configuration file, which is somewhat inconvenient. However, if these two configuration lines are placed in `load-once.conf`, they can be loaded automatically.

Additionally, if someone develops a new Waterbot, they can also use this file to include the necessary settings when releasing it.

mercury233:
> We may need two files, `system-default.conf` and `load-once.conf`
> 
> `system-default.conf` is for distro default config, and `load-once.conf` is for apply and then delete.

I think adding another configuration file is more confusing than convenient. Distributors usually need to compile the program themselves, so the default configuration can be directly written into the source code.

purerosefallen:
> If a distribution requires default settings that differ from those in the source code, it can use a `load-once.conf` file instead. The program will read this file at startup and automatically delete it afterward.

it should use `system-default.conf` instead, or the above word makes no sense at all.

purerosefallen:
note:
Will be reverted in mycard/ygopro due to lack of adaption in MyCard client.

mercury233:
> _[purerosefallen](https://github.com/purerosefallen) added the [need-eco-adapt](https://github.com/Fluorohydride/ygopro/issues?q=state%3Aopen%20label%3Aneed-eco-adapt) label [on Sep 29, 2025](https://github.com/Fluorohydride/ygopro/pull/2761#event-19972875561)_

> lack of adaption

purerosefallen:
> > _[purerosefallen](https://github.com/purerosefallen) added the [need-eco-adapt](https://github.com/Fluorohydride/ygopro/issues?q=state%3Aopen%20label%3Aneed-eco-adapt) label [on Sep 29, 2025](https://github.com/Fluorohydride/ygopro/pull/2761#event-19972875561)_
> 
> > lack of adaption

https://code.moenext.com/mycard/mycard/-/merge_requests/33

I have made the adaption in the above PR to MyCard client. What is needed is to add the configuration and publish a release containing this one, and fully test this one.

mercury233:
Why do you tell me that? Aren't you who maintain mycard/ygopro the person who should do that?

salix5:
What is the problem of this PR?

purerosefallen:
> What is the problem of this PR?

Removal of system.conf would require some adaptions in ygopro launcher / updater such as mycard

The updater would consider system.conf as a deleted file and it would delete for every user, cause users losing their profile files.

Since no one is willing to do the adaption so it should be reverted in mycard distro to avoid disasters

This is an explanation to the above discussion

salix5:
> The updater would consider system.conf as a deleted file and it would delete for every user, cause users losing their profile files.

I don't understand.
Why should the "updater" delete system.conf? 
The original system.conf are still on the local disks, and it is outside the version control.

PR #2770, opened 2025-04-27
更新字体的使用方式

yybbwc:
Changes by Michael Zeilfelder:
	- Remove ustring (otherwise I risk converting strings twice in my project as they are UCS-2 or UCS-4 already)
	- Remove const-ref in places where it makes no sense
	- Remove additional create functions and keep just one. Pass only the parameters actually needed to it.
	- Removing allocator as I get memory-leaks and don't understand what the allocater was meant for.
	- Glyph_Pages no longer mutable (doesn't seem to need that)
	- remove textscenenodes. Probably shouldn't be in this class and make understanding it harder.
	- remove createTextureFromChar (it also had a memory leak probably as it wasn't dropping pageholder)
	- remove Glyphs.set_free_when_destroyed(false). That function prevents clearing the array - it's not about pointers. Rather fix SGUITTGlyph (which didn't care about rule of three).
	- don't drop the driver (not grabbed - but I changed that stuff a lot so might not have been there originally)
	- remove a few memory re-allocations in drawing
	- CGUITTGlyphPage::updateTexture replaced dirtyflag by check for glyphs to handle
	- CGUITTGlyphPage::updateTexture works with ::texture->getSize() instead of getOriginalSize. Same result in this case, but more correct.
	- Irrlichtify code (variable naming etc)
	- Add support for outlines (lot of code got changed for that, I guess original code now barely recognizable)
	- Ensure Texture is grab()'ed. It's dangerous to remove it from driver otherwise (which might have cleared it's cache already)
	- Rename "use" functions to "get" so we have setters and getters
	- LineHeight and maximal character height now only calculated once after load
	- Using LineHeight not just for getDimension, but also for draw calls.
	- Remove unused function CGUITTFont::getDimensionUntilEndOfLine

* 我把font hint给默认打开了，这样才与之前的版本相对应
* 大概能行吧……


mercury233:
这个CGUITTFont是哪个版本，从哪里来的，是否经过实际使用，有什么好处

yybbwc:
> 这个CGUITTFont是哪个版本，从哪里来的，是否经过实际使用，有什么好处

* 最新版本
* 当前irrlicht的维护者那里。在论坛里能看到他的存储库网址
* 我在用
* 好处就是他修改的那些点……

yybbwc:
https://github.com/mzeilfelder/irr-playground-micha

mercury233:
https://github.com/mzeilfelder/irr-playground-micha/issues/1

yybbwc:
> [mzeilfelder/irr-playground-micha#1](https://github.com/mzeilfelder/irr-playground-micha/issues/1)

image

* 好像好了的样子

PR #2771, opened 2025-04-28
build irrlicht with premake on macos

mercury233:


purerosefallen:
zh99998's Mac test builder broke. I would look through after zh fix it

purerosefallen:
lgtm

PR #2713, opened 2025-03-03
add myfopen, fix setlocale

mercury233:
`std::fopen` support UTF-8 filename on windows since version 1803 (10.0.17134.0) of the Windows SDK, so we can use that.
In case of some developer using old SDK, a fallback is neccessary.

fix #2711
close #2751
close #2752

-----

For the value of `locale`, I've did some tests:

**Windows:**  
- `""`: doesn't work  
- `UTF-8`: doesn't work  
-  **`.UTF-8`: works (recommended by [official documentation](https://learn.microsoft.com/zh-cn/cpp/c-runtime-library/reference/setlocale-wsetlocale?view=msvc-170#utf-8-support))**  
- `C.UTF-8`: doesn't work  
- `zh_CN.UTF-8`: works  
- `en_US.UTF-8`: works  

**Linux:**  
- **`""`: works (recommended by [official documentation](https://linux.die.net/man/3/setlocale); can be overridden by environment variables; on my system, it resolves to `zh_CN.UTF-8`)** 
- `UTF-8`: doesn't work  
- `.UTF-8`: doesn't work  
- `C.UTF-8`: works  
- `zh_CN.UTF-8`: works  
- `en_US.UTF-8`: doesn't work  

**macOS:**  
- `""`: doesn't work  
- `UTF-8`: works  
- `.UTF-8`: doesn't work  
- **`C.UTF-8`: works (present in the output of `locale -a`)**  
- `zh_CN.UTF-8`: works  
- `en_US.UTF-8`: works  

Although `zh_CN.UTF-8` works, it may not align with multilingual goals.

PR #2786, opened 2025-05-11
use fopen in mywfopen

salix5:
use the same condition in myfopen 

@mercury233


mercury233:
`_wfopen` is standard function of windows, shouldn't be avoided

salix5:
If there is a function in the C++ standard, Microsoft-specific functions should be avoided.
ygopro used `_wfopen` because UTF-8 filename is not possible at that time.

Now we have UTF-8 filenames now, so using `_wfopen` is unnecessary.

mercury233:
I can't agree. One should make full use of platform-specific code instead of implementing it manually. This is an optimization.

salix5:
wchar_t is bad, and _wfopen is even worse.
It is a legacy code due to historical reasons, which is the opposite of optimization.

mercury233:
But it is obviously that we must work with `wchar_t`, and `_wfopen` is the official way to do this on Windows.

salix5:
No, it is not.
fopen and UTF-8 filename is the official way to open files on every modern file systems.
Windows has spent 10 years in building a modern file system.

mercury233:
I don't think the "modern file system" you said will support support `wchar_t`. But we must support `wchar_t`, otherwise we should change all irrlicht references.

salix5:
It will not and should not support wchar_t filenames.
So we can simply convert wstring to utf8 string, which is what this PR do.

mercury233:
`_wfopen` is the official way to support `wchar_t` on Windows.

salix5:
_wfopen is the legacy code for the older versions of Windows, which cannot get away from wchar_t filenames.
For those who use utf8 file systems, using _wfopen is asking for more troubles.

Microsoft might need another 10 years to remove this function, but we do not have to stand and wait.

salix5:
Windows supports utf8 filenames now, as you did in #2713.

And I am not talking about deprecated.
I am talking about conforming to standard whenever it is possible.
If there is a function in cpp standard (in this case, fopen), there is no reason using a platform dependant function (_wfopen).

mercury233:
> Windows supports utf8 filenames now
> For those who use utf8 file systems, using _wfopen is asking for more troubles.

So what is the trouble exactly?

> If there is a function in cpp standard (in this case, fopen), there is no reason using a platform dependant function (_wfopen).

It is an optimization to use `_wfopen` to avoid extra convert.

salix5:
> So what is the trouble exactly?

Writing extra code
Wiring extra code is a disadvantage by itself, especially when doing a very simple task like opening a file.

_wfopen has its own specification 
https://en.cppreference.com/w/cpp/io/c/fopen
The behavior is undefined if the mode is not one of the strings listed above. Some implementations define additional supported modes (e.g. Windows).

_wfopen has:
- its own parameters 
mode is also wchar_t

- its own mode strings
t
"rt+, ccs=UTF-8" (yes, multiple sections)

_wfopen is pretty much another different function.
If you believe that _wfopen is "official", then writing mode strings "rt+, ccs=UTF-8" should be acceptable.
It is asking for troubles.



> It is an optimization to use `_wfopen` to avoid extra convert.

Optimization means an observable improvement in performance.
If you don't open several hundred files per second, converting to utf8 filenames does not have an impact in performance.
So no, it is not an optimization.
It is just legacy code.

mercury233:
> The behavior is undefined if the mode is not one of the strings listed above. Some implementations define additional supported modes (e.g. Windows).

It means `std::fopen` behavior different on Windows too. It can't be avoided.

salix5:
> > The behavior is undefined if the mode is not one of the strings listed above. Some implementations define additional supported modes (e.g. Windows).
> 
> It means `std::fopen` behavior different on Windows too. It can't be avoided.

It can be avoided.
Microsoft CRT conforms to cpp standard in VS2015.
If we use fopen, the behaviour should be the same on all platforms.

salix5:
https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170
Most of them started from vs2015.



> In this docs, Microsoft don't say them no longer support "rt+, ccs=UTF-8" which you feel troublesome.

I am not talking about deprecated.
I am talking about conforming to standard whenever it is possible.

It is actually a simple choice.

If you choose to follow the standard 
The behavior is undefined if the mode is not one of the strings listed above.
mode "rt+, ccs=UTF-8" should not be accepted.

If you choose to follow Microsoft specific _wfopen
The extra string is specified.
mode "rt+, ccs=UTF-8" should be accepted.

My answer is clear.
Follow the standard whenever it is possible.
If there is a function in the standard, looking into any platform dependant functions is unnecessary.

mercury233:
> I am talking about conforming to standard whenever it is possible.

My answer is

> `_wfopen` is standard function of windows

> One should make full use of platform-specific code instead of implementing it manually. This is an optimization.

-----

> It is actually a simple choice.

Using `_wfopen` on Windows shouldn't even be a question.

-----

> mode "rt+, ccs=UTF-8" should be accepted.

I don't think so.
The system API offers countless features, and it's unrealistic for us to use all of them. It's normal to agree not to use certain features in order to maintain cross-platform compatibility.

salix5:
> It's normal to agree not to rely on certain features in order to maintain cross-platform compatibility.

Using platform dependant functions is relying on  platform specific features by itself.
How could one use platform dependant functions and "maintain cross-platform compatibility" at the same time?

salix5:
> > Using platform dependant functions is relying on platform specific features by itself. How could one use platform dependant functions and "maintain cross-platform compatibility" at the same time?
> 
> just use #ifdef

#ifdef is exactly the sign of failing to maintain cross-platform compatibility.
If ones really needs to use #ifdef due to historical reasons, they should use the standard functions in any possible situations.
Using platform dependant in #ifdef is acceptable only when there is no other choice, and that is what this PR does.

mercury233:
> #ifdef is exactly the sign of failing to maintain cross-platform compatibility.

I can't agree.

Before all major platforms provide the same implementation (I mean some fopen which accepting wchar_t), using #ifdef for cross-platform compatibility is perfectly normal.

salix5:
fopen should not accept wchar_t filenames because wchar_t is an example of Microsoft breaking the standard.

https://en.cppreference.com/w/cpp/language/types#Character_types
wchar_t — type for wide character representation (see wide strings). It has the same size, signedness, and alignment as one of the integer types, but is a distinct type. In practice, it is 32 bits and holds UTF-32 on Linux and many other non-Windows systems, but 16 bits and holds UTF-16 code units on Windows. The standard used to require wchar_t to be large enough to represent any supported character code point. However, such requirement cannot be fulfilled on Windows, and thus it is considered as a defect and removed.

2-bytes wchar_t can only hold a utf16 code unit, not the whole Unicode character.
The 2-bytes "wide character" is neither wide(too narrow for Unicode character) nor character (each one is just a code unit), and even Microsoft itself tries to fix this by supporting utf8 filenames.

Why should we stick to wchar_t filenames and _wfopen in 2025?

mercury233:
> fopen should not accept wchar_t filenames because wchar_t is an example of Microsoft breaking the standard.

Microsoft tried to create the standard but them fail. ([source](https://devblogs.microsoft.com/oldnewthing/20190830-00/?p=102823))

-----

Unless we make tons of modification to irrlicht (and much more libs), we must deal with `wchar_t`, and `_wfopen` is standard function of windows to do this.

salix5:
> > fopen should not accept wchar_t filenames because wchar_t is an example of Microsoft breaking the standard.
> 
> Microsoft tried to create the standard but them fail. ([source](https://devblogs.microsoft.com/oldnewthing/20190830-00/?p=102823))
> 
> Unless we make tons of modification to irrlicht (and much more libs), we must deal with `wchar_t`, and `_wfopen` is standard function of windows to do this.

We do not have modify irrlicht at all.
This PR does not modify irrlicht.

We only need to treat the irrlicht text as wchar_t *strings*, not wchar_t *filenames*.
Filenames are always utf8 strings in modern Windows versions.

mercury233:
When handling file name strings of type wchar_t, the ideal scenario is to have a function that directly supports this parameter. In less convenient cases, it is necessary to convert it into other formats, such as char, in order to be supported by available functions. 
我们没有必要舍本逐末。

salix5:
The ideal scenario is using utf8 filenames everywhere.
Most libraries except irrlicht takes utf8 filenames.
I think that is why we need #2713.

mercury233:
> The ideal scenario is using utf8 filenames everywhere

wchar_t and UTF-8 are not parallel concepts. wchar_t is simply a data type, and the idea of "converting wchar_t to UTF-8" is not accurate in itself. What does exist is the concept of "converting an array of wchar_t representing UCS-2 to a char array encoded in UTF-8." and so on.

We must handle wchar_t strings, and the best way to do this on windows is...

salix5:
> wchar_t and UTF-8 are not parallel concepts.

After Microsoft breaks the standard, 2 bytes wchar_t on Windows has stored 1 utf16 code unit for more than a decade.
The utf8 code unit is 1 byte, so no one would use 2 bytes wchar_t to store 1 bye utf8 unit.
Using char to store utf8 strings are confirmed by Microsoft in the utf8 support document.

Moreover, Windows API use wchar_t tp store utf16 strings.
wchar_t on Windows means utf16 in most cases.
They are mutually exclusive concepts on Windows.



> We must handle wchar_t filenames, and the best way to do this on windows is...

We simply get a utf16 string.
Filenames are always utf8 strings, and we only need to convert it to utf8.

mercury233:
> Filenames are always utf8 strings

No, they are not. If you want to realize that vision, it would require a lot of conversions that are likely to be unnecessary.

mercury233:
> 1 conversion per fopen call

won't realize your vision of filenames are always utf8 strings.

> if we don't open several hundred files per second.

We shouldn't assume that.

salix5:
> > 1 conversion per fopen call
> 
> won't realize your vision of filenames are always utf8 strings.

It does after this PR.

mercury233:
> > > 1 conversion per fopen call
> > 
> > 
> > won't realize your vision of filenames are always utf8 strings.
> 
> It does after this PR.

I don't think so.

yybbwc:
哈哈，保持utf8文件名，很好。看看现在热门的跨平台剪贴板基础库，都会特意采用utf8编码

mercury233:
> 哈哈，保持utf8文件名，很好。看看现在热门的跨平台剪贴板基础库，都会特意采用utf8编码

这个PR不涉及utf8文件名，更不涉及移除wchar_t，只涉及对wchar_t的文件名的处理

yybbwc:
> Windows supports utf8 filenames now, as you did in

> In practice, it is 32 bits and holds UTF-32 on Linux and many other non-Windows systems, but 16 bits and holds UTF-16 code units on Windows.


> The ideal scenario is using utf8 filenames everywhere.


> Filenames are always utf8 strings

mercury233:
> 我想這個pr主要的爭點之一就是使用utf8 文件名

唯一的争议是要不要弃用`_wfopen`，理由是它不符合C++标准。实际上它符合Windows标准，所以没有争议的必要。

mercury233:
> But why should it be closed?

I've explained this to you for a day.

salix5:
me too 
What is wrong with that?

mercury233:
@salix5 @purerosefallen Can we close this one?

yybbwc:
放着不行吗？大不了不合并呗

salix5:
我還是這個PR沒有問題不用關閉
但理念差異太大那也沒辦法
交給以後的人（如果有的話）決定罷

mercury233:
> 交給以後的人
> Version 0x1362

_man what can i cay_

salix5:
If there is a to-do-list, I definitely want to my PR in the list.
I think that is why you put your PR in the list, don't you?

mercury233:
> it should not be merged until fully discussed

It is fully discussed already, the result is it should not be merged.

mercury233:
> If there is a to-do-list, I definitely want to put my PR in the list.

Do this PR block the release of the next version?

salix5:
> > If there is a to-do-list, I definitely want to put my PR in the list.
> 
> Do this PR block the release of the next version?

Do all of the items in the list block the release of the next version?

salix5:
> > it should not be merged until fully discussed
> 
> It is fully discussed already, the result is it should not be merged.
> 
> 问题是“决定”而不是“讨论”，咱们是1：1，你可能觉得无法决定。

No, it is not.
You think it should not be merged.
I think it should be merged.
I don't think it is called "the result is it should not be merged".

mercury233:
> > > it should not be merged until fully discussed
> > 
> > 
> > It is fully discussed already, the result is it should not be merged.
> > 问题是“决定”而不是“讨论”，咱们是1：1，你可能觉得无法决定。
> 
> No, it is not. You think it should not be merged. I think it should be merged. I don't think it is called "the result is it should not be merged".
> 
> I think people would call it "no conclusion".

You can't answer most of my questions. The result had came out already, the only thing left is you admit it.

mercury233:
> > > If there is a to-do-list, I definitely want to put my PR in the list.
> > 
> > 
> > Do this PR block the release of the next version?
> 
> Do all of the items in the list block the release of the next version?

I think so.

> Unless we merge these things, I don't recommend bumping the version number.

salix5:
> > > > If there is a to-do-list, I definitely want to put my PR in the list.
> > > 
> > > 
> > > Do this PR block the release of the next version?
> > 
> > 
> > Do all of the items in the list block the release of the next version?
> 
> I think so.
> 
> > Unless we merge these things, I don't recommend bumping the version number.

Then I think this PR might block the release of the next version.
Or maybe we should just stay at 0x1031
I don't really know.

salix5:
> > > > it should not be merged until fully discussed
> > > 
> > > 
> > > It is fully discussed already, the result is it should not be merged.
> > > 问题是“决定”而不是“讨论”，咱们是1：1，你可能觉得无法决定。
> > 
> > 
> > No, it is not. You think it should not be merged. I think it should be merged. I don't think it is called "the result is it should not be merged".
> > I think people would call it "no conclusion".
> 
> You can't answer most of my questions. The result had came out already, the only thing left is you admit it.

You can't answer most of my questions either.
Both of us could say that "I won" at any time.

mercury233:
> You can't answer most of my questions either.

I have answered almost all of your questions, except for the ones where you changed the topic too far off. I’ve been trying to stay on topic, while you keep diverting it.

salix5:
No, you did not.
You did not answer:
Why should we use the platform dependent function _wfopen, when `fopen` itself is available?

Converting to utf8 string with length less than 256 utf16 code units is not even a problem.
We have to do the conversion on Linux anyway.
Does it break Linux version?

mercury233:
> Why should we use the platform dependent function _wfopen, when fopen itself is available?

> > _wfopen is standard function of windows,

> > One should make full use of platform-specific code instead of implementing it manually. This is an optimization.

> Does it break Linux version?

This is a new question. As the discussion is over, I'm too lazy to answer.

salix5:
> If there is a function in the C++ standard, Microsoft-specific functions should be avoided.
> ygopro used _wfopen because UTF-8 filename is not possible at that time.
> Now we have UTF-8 filenames, so using _wfopen is unnecessary.
>  It is a legacy code due to historical reasons, which is the opposite of optimization.

mercury233:
_Should I repost my answers again?_

salix5:
No.
I just want to say that both of us has *answered* the questions of each other.
But none of us *agreed*. 
So I would personally call it "no valid conclusion".

mercury233:
No, you did not.
You did not answer:

> One should make full use of platform-specific code instead of implementing it manually. This is an optimization.

Your answer is "`wchar_t` is bad", didn't directly answer to whether "One should make full use of platform-specific code".

After I told you the fact that we must work with `wchar_t`, you should answer me the above question.

salix5:
> No, you did not. You did not answer:
> 
> > One should make full use of platform-specific code instead of implementing it manually. This is an optimization.

This is not a question.
It is a statement, and I never agree.

One should only use platform-specific code if conforming to the current C++ standard is impossible or almost impossible.
Conforming to the standard has the highest priority, far beyond "optimization".

> 
> Your answer is "`wchar_t` is bad", didn't directly answer to whether "One should make full use of platform-specific code".
> 
> After I told you the fact that we must work with `wchar_t`, you should answer me the above question.

This is also a statement, and I never agree.
We *get* a wchar_t string.
That is all.

We can convert the wchar_t string (utf16, in this case) to a utf8 string.

I don't think it is "we must work with `wchar_t`".

mercury233:
> Conforming to the standard has the highest priority

> > _wfopen is standard function of windows

> I don't think it is "we must work with wchar_t".

But it is, whatever you think.

salix5:
The C++ standard has higher priority than any 

> > Conforming to the standard has the highest priority
> 
> > > _wfopen is standard function of windows


No, it is not.
The standard function is fopen.

C99
7.19.5.3 The fopen function
Synopsis
```c
#include <stdio.h>
FILE *fopen(const char * restrict filename,
const char * restrict mode);
```
Description
The fopen function opens the file whose name is the string pointed to by filename, and associates a stream with it.

The argument mode points to a string. If the string is one of the following, the file is open in the indicated mode. 
Otherwise, the behavior is undefined.

r open text file for reading
w truncate to zero length or create text file for writing
a append; open or create text file for writing at end-of-file
rb open binary file for reading
wb truncate to zero length or create binary file for writing
ab append; open or create binary file for writing at end-of-file
r+ open text file for update (reading and writing)
w+ truncate to zero length or create text file for update
a+ append; open or create text file for update, writing at end-of-file
r+b or rb+ open binary file for update (reading and writing)
w+b or wb+ truncate to zero length or create binary file for update
a+b or ab+ append; open or create binary file for update, writing at end-of-file

r+b or rb+ open binary file for update (reading and writing)
w+b or wb+ truncate to zero length or create binary file for update
a+b or ab+ append; open or create binary file for update, writing at end-of-file


_wfopen is another different function.
It is not the standard function.
People are forced to use the non-standard function on Windows because Microsoft *breaks* the standard by creating 2-byte wchar_t.

_wfopen is not a standard function.
It is not a standard function from the beginning.
Even the name started with `_` is a indication of non-standard function.


> 
> _another question/statement you didn't answer_
> 
> > I don't think it is "we must work with wchar_t".
> 
> But it is, whatever you think.

But it is not, whatever you think.
By the way, I do not think this kind of statement could be called "answer".

mercury233:
_finally you answered that_

On Windows, fopen and _wfopen are considered equivalent in status.
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen

If _wfopen is not standard on Windows, fopen is not too.

-----

> People are forced to use the non-standard function on Windows because Microsoft _breaks_ the standard by creating 2-byte wchar_t.

Microsoft tried to create the standard but them fail. ([source](https://devblogs.microsoft.com/oldnewthing/20190830-00/?p=102823)). `wchar_t` is added to the Windows standard even before the C standard.

> The Windows classic format came first, so the question is whether the C standard chose to align with the Windows classic format, rather than vice versa.

salix5:
For the health of all participants, any question beyond 23:00 in GMT+8 timezone might be replied next day.
I will reply it tomorrow.

salix5:
> _finally you answered that_
> 
> On Windows, fopen and _wfopen are considered equivalent in status. https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen
> 
> If _wfopen is not standard on Windows, fopen is not too.

I did not see how to reach this conclusion.

I am talking about C++14 standard, where most C functions came from C99 standard.
I am 100% sure that `_wfopen` is NOT in the C++14 standard or C99 standard.

My statement is quite simple:
Can we open a file using **only** the file I/O functions in C++14/C99 standard?
Yes, we can, in the current version of Windows 10 SDK.

```c
#include <stdio.h>
FILE *fopen(const char * restrict filename,
const char * restrict mode);
```
By setting the locale to utf8, we can use `fopen` to open a file.
If we can do that, then we should do that.

> 
> > People are forced to use the non-standard function on Windows because Microsoft _breaks_ the standard by creating 2-byte wchar_t.
> 
> Microsoft tried to create the standard but them fail. ([source](https://devblogs.microsoft.com/oldnewthing/20190830-00/?p=102823)). `wchar_t` is added to the Windows standard even before the C standard.
> 
> > The Windows classic format came first, so the question is whether the C standard chose to align with the Windows classic format, rather than vice versa.

Yes, I know that the development of Windows is much earlier than C99 standard.
They chose UTF-16, and it is probably not a good choice from the viewpoint of 2025.
They also chose that each `wchar_t` will store 1 utf16 code unit, which maximize the problem of surrogate in utf16.
I am very grateful that the C99 standard does not use that definition.

But actually I have no interest about what Unicode was in days before I was born.
I only care what Unicode is in 2025.
In 2025, most file systems use utf8 filenames.

utf8 string:
- compatible with ASCII strings
- compatible with most of the C-string functions (strcmp, strncat, ...)
- 1 byte code unit, compatible with 1 byte `char` in C/C++
- `char` is 1 byte on all platform

There is `char8_t` in C++20, but that is in the future we might never reach though.
We cannot accept C++17 yet.

on the contrast, wchar_t:
- different size on different platform (Windows: 2 bytes, other: 4bytes)
- different encoding on different platform (Windows: utf16 code unit, UTF-32 characters)
- not compatible with ASCII strings
- Windows uses wchar_t to store utf16 string

If different encoding on different platform is not bad enough, utf16 has more surprises to make it even worse:
surrogate pair and lone surrogate

Even Microsoft itself tries to support utf8 filenames now.
I cannot see why we should stick to the utf16 filenames.

yybbwc:
> > _finally you answered that_
> > 
> > On Windows, fopen and _wfopen are considered equivalent in status. https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen
> > 
> > If _wfopen is not standard on Windows, fopen is not too.
> 
> I did not see how to reach this conclusion.
> 
> I am talking about C++14 standard, where most C functions came from C99 standard.
> I am 100% sure that `_wfopen` is NOT in the C++14 standard or C99 standard.
> 
> My statement is quite simple:
> Can we open a file using **only** the file I/O functions in C++14/C99 standard?
> Yes, we can, in the current version of Windows 10 SDK.
> 
> ```c
> #include <stdio.h>
> FILE *fopen(const char * restrict filename,
> const char * restrict mode);
> ```
> By setting the locale to utf8, we can use `fopen` to open a file.
> If we can do that, then we should do that.
> 
> > 
> > > People are forced to use the non-standard function on Windows because Microsoft _breaks_ the standard by creating 2-byte wchar_t.
> > 
> > Microsoft tried to create the standard but them fail. ([source](https://devblogs.microsoft.com/oldnewthing/20190830-00/?p=102823)). `wchar_t` is added to the Windows standard even before the C standard.
> > 
> > > The Windows classic format came first, so the question is whether the C standard chose to align with the Windows classic format, rather than vice versa.
> 
> Yes, I know that the development of Windows is much earlier than C99 standard.
> They chose UTF-16, and it is probably not a good choice from the viewpoint of 2025.
> They also chose that each `wchar_t` will store 1 utf16 code unit, which maximize the problem of surrogate in utf16.
> I am very grateful that the C99 standard does not use that definition.
> 
> But actually I have no interest about what Unicode was in days before I was born.
> I only care about what Unicode is in 2025.
> In 2025, most file systems use utf8 filenames.
> 
> utf8 string:
> - compatible with ASCII strings
> - compatible with most of the C-string functions (strcmp, strncat, ...)
> - 1 byte code unit, compatible with 1 byte `char` in C/C++
> - `char` is 1 byte on all platform
> 
> There is `char8_t` in C++20, but that is in the future we might never reach though.
> We cannot accept C++17 yet.
> 
> on the contrast, wchar_t:
> - different size on different platform (Windows: 2 bytes, other: 4bytes)
> - different encoding on different platform (Windows: utf16 code unit, UTF-32 characters)
> - not compatible with ASCII strings
> - Windows uses wchar_t to store utf16 string
> 
> If different encoding on different platform is not bad enough, utf16 has more surprises to make it even worse:
> surrogate pair and lone surrogate
> 
> Even Microsoft itself tries to support utf8 filenames now.
> I cannot see why we should stick to the utf16 filenames.
> 

顺便补充一点。
 wchar_t还有个缺点，就是字节序。

mercury233:
What you say only leads to we shouldn't use `wchar_t`, but the problem we are discussing is how to use `wchar_t`, we can't just don't use `wchar_t`.

> Can we open a file using only the file I/O functions in C++14/C99 standard?

Surely we can, but why should we do that? We are using MSVC on Windows, it is better to follow Microsoft standard.

salix5:
> What you say only leads to we shouldn't use `wchar_t`, but the problem we are discussing is how to use `wchar_t`, we can't just don't use `wchar_t`.

We shouldn't use `wchar_t` if possible.
Even if we get a `wchar_t` string from Irrlicht, we can convert it to a utf8 string.
We can try to avoid `wchar_t` as much as possible.

> 
> > Can we open a file using only the file I/O functions in C++14/C99 standard?
> 
> Surely we can, but why should we do that? We are using MSVC on Windows, it is better to follow Microsoft standard.

It is better to follow C++14/C99 standard whenever it is possible, even if we use MSVC on Windows.
MSVC has lots of Microsoft-specific functions, and even more surprises in the specifications.

It is a problem of priority.
If Microsoft-specific functions take a higher priority just because we are on Windows , we are actually embracing all the surprises.
It might end up with writing more useless codes outside Windows.

mercury233:
> Even if we get a `wchar_t` string from Irrlicht, we can convert it to a utf8 string.

“convert” itself need to "use" `wchar_t` param, same as wfopen.

> It is better to follow C++14/C99 standard whenever it is possible, even if we use MSVC on Windows.

You can't actually use the "C++14/C99 standard" `std::fopen` on MSVC, you can only use MSVC `fopen` on MSVC.

> It is a problem of priority.

Beside for the standard, using 1 `_wfopen` is better than 1 conversion + 1 `fopen` in performance.

salix5:
> > Even if we get a `wchar_t` string from Irrlicht, we can convert it to a utf8 string.
> 
> “convert” itself need to "use" `wchar_t` param, same as wfopen.

You know it is nitpicking.
I'm talking about avoiding wchar_t *filenames* all alone.
So I have to say that again: 

My statement is quite simple:
Can we open a file using only the file I/O functions in C++14/C99 standard?
Yes, we can, in the current version of Windows 10 SDK.

By setting the locale to utf8, we can use fopen to open a file.
If we can do that, then we should do that.

> 
> > MSVC has lots of Microsoft-specific functions, and even more surprises in the specifications.
> 
> You can't actually use the "C++14/C99 standard" `std::fopen` on MSVC, you can only use MSVC `fopen` on MSVC.

MSVC `fopen` is a good approximation, and it is way better than Microsoft specific functions.


> 
> > It is a problem of priority.
> 
> Beside for the standard, using 1 `_wfopen` is better than 1 conversion + 1 `fopen` in performance.

This us the

salix5:
> 顺便补充一点。 wchar_t还有个缺点，就是字节序。

謝謝
我還忘了提endianness, BOM
以及微軟曾經選擇在utf8 檔案加入BOM創造更多問題…

mercury233:
> I'm talking about avoiding wchar_t _filenames_ all alone.

What we passed to `mywfopen` must be filename.

> Can we open a file using only the file I/O functions in C++14/C99 standard?

Is this a question from a interview? I can't see any relation from this question to real coding.

> MSVC fopen is a good approximation, and it is way better than Microsoft specific functions.

Why? It took the same params and return value from _wfopen.

> If you don't open several hundred files per second

Do we really have to make this assumption? If so, we must keep in mind that we can't open multiple files simultaneously when using mywfopen, otherwise there will be a performance drop on Windows. That seems more like the kind of "surprise" you mentioned.

_By the way, nearly a hundred scripts from both players' decks are indeed loaded simultaneously at the start of a duel._

salix5:
> > I'm talking about avoiding wchar_t _filenames_ all alone.
> 
> What we passed to `mywfopen` must be filename.

It can be converted to utf8 filenames.

> 
> > Can we open a file using only the file I/O functions in C++14/C99 standard?
> 
> Is this a question from a interview? I can't see any relation from this question to real coding.

I am talking about this from the first comment.
Conforming to the standard.
Using utf8 filenames.
They are the same thing.

> 
> > MSVC fopen is a good approximation, and it is way better than Microsoft specific functions.
> 
> Why? It took the same params and return value from _wfopen.

No
_wfopen does not take the same params.
_wfopen takes wchat_t

> 
> > If you don't open several hundred files per second
> 
> Do we really have to make this assumption? If so, we must keep in mind that we can't open multiple files simultaneously when using mywfopen, otherwise there will be a performance drop on Windows. That seems more like the kind of "surprise" you mentioned.
> 
> _By the way, nearly a hundred scripts from both players' decks are indeed loaded simultaneously at the start of a duel._

You still did not answer:
We have to convert to utf8 on Linux or any other file systems.
Does it break the performance on Linux?
I think not.

mercury233:
> It can be converted to utf8 filenames.

There’s no need to deceive ourselves. The input is indeed a file name. Using it in `_wfopen` and using it in `wcsrtombs` are both uses.

> I am talking about this from the first comment.

I am talking about this from the first comment too.

>> `_wfopen` is standard function of windows, shouldn't be avoided

> Conforming to the standard.
Using utf8 filenames.
They are the same thing.

Conforming to the standard.
Using `_wfopen` on Windows.
They are the same thing.

> We have to convert to utf8 on Linux or any other file systems. Does it break the performance on Linux?

You said. We have to. But on Windows we don't have to. It do make the performance very slightly lower on Linux.

salix5:
If we are going to repeat our viewpoint for 114514 times, then it is time to stop.
As I said before, no valid conclusion.

PR #2790, opened 2025-05-16
display side count to removed

purerosefallen:
image


mercury233:
I guess this is adding the intelligence previously missing, and only appear when selecting to go first or second?

purerosefallen:
> I guess this is adding the intelligence previously missing, and only appear when selecting to go first or second?

whenever the main and extra displays. when duel starts it would not appear.

purerosefallen:
anything left?

purerosefallen:
it's also a Chocolete-test-passed thing. could we do this one now?
@mercury233

PR #2799, opened 2025-05-19
fix CopyString & CopyWideString

mercury233:
Using strncat requires that:

1. src must be null-terminated.
2. dst must be a valid C string (null-terminated and properly initialized).

If src is not null-terminated, or dst contains old data, strncat can cause undefined behavior, like memory corruption or crashes.

Using strncpy is safer:

* It copies a fixed number of characters.
* You can manually add the null terminator.
* It does not depend on the existing content of dst.

### DIFF
--- a/gframe/bufferio.h
+++ b/gframe/bufferio.h
@@ -62,13 +62,13 @@ public:
 	}
 	template<size_t N>
 	static void CopyString(const char* src, char(&dst)[N]) {
-		dst[0] = 0;
-		std::strncat(dst, src, N - 1);
+		std::strncpy(dst, src, N - 1);
+		dst[N - 1] = 0;
 	}
 	template<size_t N>
 	static void CopyWideString(const wchar_t* src, wchar_t(&dst)[N]) {
-		dst[0] = 0;
-		std::wcsncat(dst, src, N - 1);
+		std::wcsncpy(dst, src, N - 1);
+		dst[N - 1] = 0;
 	}
 	template<typename T>
 	static bool CheckUTF8Byte(const T* str, int len) {


salix5:
> Using strncat requires that:
> 
> 1. src must be null-terminated.

all c string functions requires null terminated string 

> 2. dst must be a valid C string (null-terminated and properly initialized).

dst[0] = 0;
dst is a valid string

> 
> If src is not null-terminated, or dst contains old data, strncat can cause undefined behavior, like memory corruption or crashes.
> 
> Using strncpy is safer:
> 
> * It copies a fixed number of characters.

strncat does too

> * You can manually add the null terminator.

strncat will add it by itself 
strncpy is not designed for null terminated string.
That's why it has an odd behaviour (does not add \0 in some cases)

> * It does not depend on the existing content of dst.

dst is an empty string now
It does not depend on the existing content of dst

mercury233:
It is always better to be more safe.

mercury233:
Actually I got some errors in Valgrind:

```
==14734== Invalid read of size 32
==14734==    at 0x5258F43: __wcsncat_avx2 (strncat-avx2.S:80)
==14734==    by 0x2484FA: CopyWideString<20> (bufferio.h:71)
==14734==    by 0x2484FA: ygo::ReplayMode::StartDuel() (replay_mode.cpp:169)
```

I don't know it is a bug of our code, or Valgrind, or __wcsncat_avx2, but using strncpy is safer, easier to read by human and compiler.

salix5:
> It is always better to be more safe.

*Calculating* the position of \0 by human would most likely make it unsafe.
strncat approach always put \0 at the beginning of dst.

By the way, the original functions are correct.
So it is not actually a "fix".

mercury233:
> Calculating the position of \0 by human would most likely make it unsafe.

Do you mean requiring the src has correct \0 is unsafe? `strncpy` is actually setting \0 automatically.

salix5:
C99
7.21.2.4 The strncpy function
If the array pointed to by s2 is a string that is shorter than n characters, null characters
 are appended to the copy in the array pointed to by s1, until n characters in all have been
 written.

Please check the standard first.

mercury233:
???
I mean "null characters are appended to the copy in the array pointed to by s1", what do you mean?

salix5:
```c
#include <string.h>
char *strncpy(char * restrict s1,
const char * restrict s2,
size_t n);
```

The strncpy function copies not more than n characters (characters that follow a null character are not copied) from the array pointed to by s2 to the array pointed to by s1.
If copying takes place between objects that overlap, the behavior is undefined.

If the array pointed to by s2 is a string that is shorter than n characters, null characters are appended to the copy in the array pointed to by s1, until n characters in all have been written.

It means
If the array pointed to by s2 is a string that is shorter than n characters
add null character 

If not
does not add null character 

Check the standard, please.

mercury233:
???

So the "strncpy N-1, set last 0" way always make sure the dst is ended with 0, isn't it?

And the "set first 0, strncat N-1" way may fail to set 0 if the src isn't ended with 0, isn't it?

mercury233:
The param of strncat and strncpy is char*, it is not string.

mercury233:
If you think a tool function should wrong out if the input is wrong, I can close this.

mercury233:
> If you think a tool function should wrong out if the input is wrong, I can close this. And I suggest you to remove all safe features too.

salix5:
The C string functions manipulate strings, which is not very surprising.
If one developer pass a char array without null character to the C string functions, then he is bad and unsafe, not the C string functions.

But strncat is so kind that it is willing to give this bad and unsafe guy an opportunity:

```c
#include <string.h>
char *strncat(char * restrict s1, const char * restrict s2, size_t n);
```
The `strncat` function appends not more than n characters (a null character and characters that follow it are not appended) from the array pointed to by s2 to the end of the string pointed to by s1.
The initial character of s2 overwrites the null character at the end of s1. 
A terminating null character is always appended to the result.

As long as n characters does not go beyond the boundaries of s2, strncat will append n characters to s1 and add a null character in s1 in all cases.

On the contrary, strncpy will not add a null  character in some cases.
That is why you have to add a null character manually.

Of course a function that *always* add a null character is much safer than a function that *sometimes* add a null character.




> If you think a tool function should wrong out if the input is wrong, I can close this.

I suggest you close it right now, unless you are willing to read the C99 standard on your own.
All the answers yo your questions can be found in C99 standard.
Please download N1256, read it, and tell me which question you cannot find an answer in it.

mercury233:
Sorry, I misunderstood `strncat`. Because the standard mentions `strlen`, I thought it would always perform an operation equivalent to `strlen`.

It now seems that the issue I encountered may stem from an incorrect implementation of `__wcsncat_avx2`.

However, in fact, neither the original implementation nor the one proposed in this PR can properly handle cases where `src` is shorter than `N-1` but not correctly null-terminated. Both of them will read beyond `src`.

mercury233:
[log.zip](https://github.com/user-attachments/files/20298381/log.zip)

and another that may happen:

```
==9573== Thread 1:
==9573== Invalid read of size 32
==9573==    at 0x5258F43: __wcsncat_avx2 (strncat-avx2.S:80)
==9573==    by 0x220FE9: CopyWideString<20> (bufferio.h:71)
==9573==    by 0x220FE9: ygo::Game::SaveConfig() (game.cpp:1510)
==9573==    by 0x22A7F2: ygo::Game::MainLoop() (game.cpp:1105)
==9573==    by 0x199A7B: main (gframe.cpp:205)
==9573==  Address 0x158a9ad0 is 0 bytes inside a block of size 28 alloc'd
==9573==    at 0x4846FA3: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==9573==    by 0x579DE2: internal_new (irrAllocator.h:60)
==9573==    by 0x579DE2: allocate (irrAllocator.h:35)
==9573==    by 0x579DE2: operator=<wchar_t> (irrString.h:374)
==9573==    by 0x579DE2: irr::gui::CGUIEditBox::CGUIEditBox(wchar_t const*, bool, irr::gui::IGUIEnvironment*, irr::gui::IGUIElement*, int, irr::core::rect<int> const&) (CGUIEditBox.cpp:46)
==9573==    by 0x42E60D: irr::gui::CGUIEnvironment::addEditBox(wchar_t const*, irr::core::rect<int> const&, bool, irr::gui::IGUIElement*, int) (CGUIEnvironment.cpp:1270)
==9573==    by 0x22E208: ygo::Game::Initialize() (game.cpp:237)
==9573==    by 0x199572: main (gframe.cpp:70)
==9573== 
```

_And the replay in the zip contains another bug, some effects in `Auxiliary.merge_single_effects` are reset by a hack implement of Drake Shark._

salix5:
@mercury233 
Do you think we should use
```cpp
char16_t players[4][20];			// 40 bytes per player
size_t player_count{};
```

salix5:
copilot:
However, some optimized implementations (especially with SIMD/vectorization) may read in chunks (e.g., 16 or 32 bytes) past the end of the string, searching for a null.
This means: Even if it only copies n characters, it might read past src + n while searching for a null.

salix5 (approved):

salix5:
Consider SIMD or any other optimization, strncpy might be better.
Thank you.

mercury233:
https://chatgpt.com/share/682bd365-1190-800d-aa63-11741d90a1dc

purerosefallen:
is the coversation done, or is it mergable?

purerosefallen:
> > is the coversation done, or is it mergable?
> 
> Didn't fix the "not null-terminated short src" problem, but it can be merged.

is it creating a new problem, or it failed to solve a previous problem?

is the updated function a drop-in replacement to the old function?

mercury233:
> is it creating a new problem, or it failed to solve a previous problem?
> 
> is the updated function a drop-in replacement to the old function?

It is the questions who review this PR should answer.

purerosefallen:
> > is it creating a new problem, or it failed to solve a previous problem?
> > is the updated function a drop-in replacement to the old function?
> 
> Them are the questions who review this PR should answer.

In short, yes or no: can we merge this?

salix5:
> > > is the coversation done, or is it mergable?
> > 
> > 
> > Didn't fix the "not null-terminated short src" problem, but it can be merged.
> 
> is it creating a new problem, or it failed to solve a previous problem?

strncat has to find the \0 in dest, which might be a problem sometimes.
Now I agree that strncpy is better.

> 
> is the updated function a drop-in replacement to the old function?

Yes

mercury233:
> > > is it creating a new problem, or it failed to solve a previous problem?
> > > is the updated function a drop-in replacement to the old function?
> > 
> > 
> > Them are the questions who review this PR should answer.
> 
> In short, yes or no: can we merge this?

_Now it is your turn to answer this_

PR #2802, opened 2025-05-20
 add MenuHandler::temp_replay 

salix5:
ReplayMode::cur_replay
It is mainly used in ReplayThread(), so we should use another object in the main thread.

@mercury233


purerosefallen:
lgtm

PR #2807, opened 2025-05-22
 check header in OpenReplay 

salix5:
-  check header in OpenReplay 
- remove Replay::CheckReplay
`CheckReplay` cannot check the detail of yrp file.
Now we can check the yrp file by `OpenReplay`.

- does not read yrp file in RefreshReplay
Opening and reading every files takes too much time and has very limited effects.
`OpenReplay` will check the file anyway.

- add version_info for old replay 
yrp file of version before 0x1353 has many problems in PRNG, and these file need additional attention.

@mercury233 
@Wind2009-Louse


purerosefallen:
lgtm

PR #2809, opened 2025-05-23
fix signed/unsigned compare

salix5:


mercury233:
> image_manager.cpp(160,15): error C3016: 'dy': index variable in OpenMP 'for' statement must have signed integral type

_strange limit in openmp_

mercury233:
> Starting in Visual Studio 2019 version 16.9, you can use the experimental /openmp:llvm option instead of /openmp to target the LLVM OpenMP runtime. Support currently isn't available for production code, since the required libomp DLLs aren't redistributable.
https://learn.microsoft.com/en-us/cpp/build/reference/openmp-enable-openmp-2-0-support?view=msvc-170

salix5:
> > Starting in Visual Studio 2019 version 16.9, you can use the experimental /openmp:llvm option instead of /openmp to target the LLVM OpenMP runtime. Support currently isn't available for production code, since the required libomp DLLs aren't redistributable.
> > https://learn.microsoft.com/en-us/cpp/build/reference/openmp-enable-openmp-2-0-support?view=msvc-170

Yes I know.
Do you think we should not use /openmp:llvm?

salix5:
give up
love openmp

PR #2812, opened 2025-05-25
sqlite: add compile-time options

salix5:
https://sqlite.org/compile.html

[SQLITE_DQS=0](https://sqlite.org/compile.html#dqs). 
This setting disables the [double-quoted string literal](https://sqlite.org/quirks.html#dblquote) misfeature.

[SQLITE_DEFAULT_MEMSTATUS=0](https://sqlite.org/compile.html#default_memstatus). 
This setting causes the [sqlite3_status()](https://sqlite.org/c3ref/status.html) interfaces that track memory usage to be disabled. This helps the [sqlite3_malloc()](https://sqlite.org/c3ref/free.html) routines run much faster, and since SQLite uses [sqlite3_malloc()](https://sqlite.org/c3ref/free.html) internally, this helps to make the entire library faster.

[SQLITE_MAX_EXPR_DEPTH=0](https://sqlite.org/limits.html#max_expr_depth). 
Setting the maximum expression parse-tree depth to zero disables all checking of the expression parse-tree depth, which simplifies the code resulting in faster execution, and helps the parse tree to use less memory.

[SQLITE_OMIT_DECLTYPE](https://sqlite.org/compile.html#omit_decltype). 
By omitting the (seldom-needed) ability to return the declared type of columns from the result set of query, [prepared statements](https://sqlite.org/c3ref/stmt.html) can be made to consume less memory.

[SQLITE_OMIT_DEPRECATED](https://sqlite.org/compile.html#omit_deprecated). 
Omitting deprecated interfaces and features will not help SQLite to run any faster. It will reduce the library footprint, however. And it is the right thing to do.

[SQLITE_OMIT_PROGRESS_CALLBACK](https://sqlite.org/compile.html#omit_progress_callback). 
The progress handler callback counter must be checked in the inner loop of the [bytecode engine](https://sqlite.org/opcode.html). By omitting this interface, a single conditional is removed from the inner loop of the [bytecode engine](https://sqlite.org/opcode.html), helping SQL statements to run slightly faster.

[SQLITE_OMIT_SHARED_CACHE](https://sqlite.org/compile.html#omit_shared_cache). 
Omitting the possibility of using [shared cache](https://sqlite.org/sharedcache.html) allows many conditionals in performance-critical sections of the code to be eliminated. This can give a noticeable improvement in performance.

@mercury233 
@purerosefallen 
@Wind2009-Louse
@fallenstardust

### DIFF
--- a/premake/sqlite3/premake5.lua
+++ b/premake/sqlite3/premake5.lua
@@ -2,3 +2,12 @@ project "sqlite3"
     kind "StaticLib"
 
     files { "sqlite3.c", "sqlite3.h" }
+    defines {
+        "SQLITE_DQS=0",
+        "SQLITE_DEFAULT_MEMSTATUS=0",
+        "SQLITE_MAX_EXPR_DEPTH=0",
+        "SQLITE_OMIT_DECLTYPE",
+        "SQLITE_OMIT_DEPRECATED",
+        "SQLITE_OMIT_PROGRESS_CALLBACK",
+        "SQLITE_OMIT_SHARED_CACHE",
+    }


purerosefallen:
Will test it along with windows and static build premake thing

purerosefallen (approved):

PR #2814, opened 2025-05-31
always open conformance mode

salix5:
https://learn.microsoft.com/en-us/cpp/build/reference/permissive-standards-conformance?view=msvc-170

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -337,9 +337,7 @@ workspace "YGOPro"
 
     filter "action:vs*"
         cdialect "C11"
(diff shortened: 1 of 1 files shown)


mercury233:
do toolset v141_xp support this?

salix5:
I am not sure.

mercury233:
https://ci.appveyor.com/project/mercury233/ygopro/builds/52156187

It does not.

Since we have already checked conformance on other versions of the toolset, this check can be exempted in order to maintain support for XP.

PR #2822, opened 2025-06-01
add config for image load background thread and multi thread

mercury233:
* Some users have very slow CPUs or hard drives, and loading images on the main thread can cause noticeable stuttering. They may prefer to revert to the previous method, which loads images in a separate thread, even if it means giving up the optimization that hides the loading process.
* A very small number of users have experienced issues such as crashes or blue screens when using OpenMP. These users may try disabling OpenMP as a workaround.


purerosefallen:
plz add a startup param `--openmp 0` `--openmp 1`.
I think we could detect whether openmp in MyCard client or something like that.

purerosefallen:
https://github.com/Fluorohydride/ygopro/pull/2761

btw without this change, should we prepare different default system.conf for mac?

purerosefallen:
anything left?

purerosefallen:
it's a test-passed thing according to the Chocolete thing, merging it

PR #2815, opened 2025-05-31
cleanup premake links

mercury233:
* The `links` directive in Premake applies to a project, not to the workspace.
* Specifying `links` for a static library has no actual effect.
* On Windows, libraries such as `opengl32` and `imm32` are already brought in via `#pragma comment(lib, ...)`.
* On Windows, system libraries like `kernel32` and `user32` are linked automatically by MSVC.
* Starting with [glibc 2.34](https://sourceware.org/pipermail/libc-alpha/2021-August/129718.html), there is no longer any need to link against `libdl` or `libpthread` explicitly.
* `libz` (zlib) is bundled in the source trees of Irrlicht and FreeType, so it does not need to be linked separately.
* `libm` (`math.h`) is a dependency of libstdc++, so it also does not need to be linked explicitly.

PR #2823, opened 2025-06-03
premake: add glibc detect and add dl back

purerosefallen:
https://github.com/Fluorohydride/ygopro/pull/2815

broke on glibc < 2.34, so we have to detect glibc version

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -260,20 +260,23 @@ end
 
 function getGlibcVersion()
     local output = os.outputof("getconf GNU_LIBC_VERSION")
-    local major, minor = output:match("glibc (%d+)%.(%d+)")
+    local major, minor, patch = output:match("glibc (%d+)%.(%d+)%.?(%d*)")
+
     if major and minor then
         major = tonumber(major)
         minor = tonumber(minor)
-        return (major << 8) | minor
+        patch = tonumber(patch) or 0
+        return (major << 16) | (minor << 8) | patch
     end
-    return 0
+
+    return nil
 end
 
 GLIBC_VERSION=0
 if os.ishost("linux") then
     GLIBC_VERSION = getGlibcVersion()
     if GLIBC_VERSION>0 then
-        print("Detected glibc version: " .. string.format("%d.%d", GLIBC_VERSION >> 8, GLIBC_VERSION & 0xFF))
+        print("Detected glibc version: " .. string.format("%d.%d.%d", GLIBC_VERSION >> 16, (GLIBC_VERSION >> 8) & 0xFF, GLIBC_VERSION & 0xFF))
     else
         print("Could not detect glibc version, assuming it is sufficient.")
     end
--- a/premake5.lua
+++ b/premake5.lua
@@ -258,6 +258,27 @@ if os.istarget("macosx") then
     end
 end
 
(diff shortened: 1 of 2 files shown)


mercury233:
I prefer the easy way since newer glibc provide empty libpthread.a and so on.

purerosefallen:
> I prefer the easy way since newer glibc provide empty libpthread.a and so on.

always make `ldd ygopro` clean

premake5.lua
```diff
@@ -258,6 +258,30 @@ if os.istarget("macosx") then
...
+    return nil
+end
+
+GLIBC_VERSION=0
+if os.ishost("linux") then
+    GLIBC_VERSION = getGlibcVersion()
+    if GLIBC_VERSION>0 then
+        print("Detected glibc version: " .. string.format("%d.%d.%d", GLIBC_VERSION >> 16, (GLIBC_VERSION >> 8) & 0xFF, GLIBC_VERSION & 0xFF))
+    else
+        print("Could not detect glibc version, assuming it is sufficient.")
```
mercury233: insufficient?

premake5.lua
```diff
@@ -258,6 +258,30 @@ if os.istarget("macosx") then
...
+    local major, minor, patch = output:match("glibc (%d+)%.(%d+)%.?(%d*)")
+
+    if major and minor then
+        major = tonumber(major)
+        minor = tonumber(minor)
+        patch = tonumber(patch) or 0
+        return (major << 16) | (minor << 8) | patch
+    end
+
+    return nil
```
mercury233: 0?

premake5.lua
```diff
@@ -258,6 +258,30 @@ if os.istarget("macosx") then
...
+end
+
+GLIBC_VERSION=0
+if os.ishost("linux") then
```
mercury233, line 276: istarget?

premake5.lua
```diff
@@ -258,6 +258,30 @@ if os.istarget("macosx") then
...
+    GLIBC_VERSION = getGlibcVersion()
+    if GLIBC_VERSION>0 then
+        print("Detected glibc version: " .. string.format("%d.%d.%d", GLIBC_VERSION >> 16, (GLIBC_VERSION >> 8) & 0xFF, GLIBC_VERSION & 0xFF))
```
mercury233, line 279: `2.34.0` is not actual version, it should be `2.34`. btw I don't think it is necessary to print this.

mercury233 (changes requested):

premake5.lua
purerosefallen: `istarget` may not work, because there may not be `getconf GNU_LIBC_VERSION` on non-Linux OSes. WIndows / Mac generating Linux would still not call this method.

if Linux generating other OSes, this function still works.

premake5.lua
purerosefallen: Not interested in detailed format, and this is for debugging only, just as the Rosetta thing.

mercury233:
> > I prefer the easy way since newer glibc provide empty libpthread.a and so on.
> 
> always make `ldd ygopro` clean

You can do that when you distribute YGOPro. I think it is better to keep the code clean.

PR #2828, opened 2025-06-08
Do with MSG_REVERSE_DECK in reconnect

purerosefallen:
this previously didn't work because MSG_REVERSE_DECK wasn't detected.


purerosefallen:
This was intended in server PR, but opened by an error

PR #2829, opened 2025-06-08
add back link pthread

mercury233:
removed in https://github.com/Fluorohydride/ygopro/pull/2815

Some Linux distributions still support glibc versions earlier than 2.34, so we should support them as well.

As of now, linking to pthread on glibc 2.34 and later doesn't appear to have any side effects. It may take a few more years before glibc stops providing an empty libpthread.

### DIFF
--- a/gframe/premake5.lua
+++ b/gframe/premake5.lua
@@ -103,7 +103,7 @@ project "YGOPro"
         end
 
     filter "system:linux"
-        links { "GL", "X11", "Xxf86vm" }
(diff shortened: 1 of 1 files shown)


purerosefallen (approved):

purerosefallen:
side note: MyCard build YGOPro in Debian 11, glibc 2.31
This is the version supporting mostly all current LTS Linux distros.

purerosefallen:
PLEASE DO NOT EDIT MY WORDS

you may say *that's not we but mycard* below, but edit other's words is nothing acceptable.

PR #2830, opened 2025-06-08
check packet size by static_assert

salix5:


purerosefallen:
What would happen if it becomes too large?
Could you offer me an example of bad code, and how it compiles?

salix5:
> What would happen if it becomes too large? Could you offer me an example of bad code, and how it compiles?

You will get an error like:
> static assertion failed with "Packet size is too large."

So you cannot compile.
You can see this error if you change the condition to:
```cpp
static_assert(sizeof(ST) > MAX_DATA_SIZE, "Packet size is too large.");
```

MAX_DATA_SIZE is 65534 now.
If a struct is bigger than this size, it should not be compiled.

PR #2831, opened 2025-06-08
premake: add -static-libstdc++ and -static-libgcc

purerosefallen:
make less entries in `ldd` and add compatitability

MyCard builds like this for years

### DIFF
--- a/gframe/premake5.lua
+++ b/gframe/premake5.lua
@@ -104,7 +104,7 @@ project "YGOPro"
 
     filter "system:linux"
         links { "GL", "X11", "Xxf86vm" }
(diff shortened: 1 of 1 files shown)


mercury233:
> linux-vdso.so.1
libGL.so.1
libX11.so.6
libXxf86vm.so.1
libm.so.6
libgomp.so.1
libc.so.6
/lib64/ld-linux-x86-64.so.2
libxcb.so.1
libXau.so.6
libXdmcp.so.6
libbsd.so.0
libmd.so.0

not very useful

purerosefallen:
boring

PR #2833, opened 2025-06-09
patch miniaudio

mercury233:
Problem:
[Crash when using wetype IME (微信输入法), Use-After-Free when using sogou IME (搜狗输入法)](https://github.com/mackron/miniaudio/issues/996)

Solution:
Apply https://github.com/mackron/miniaudio/pull/968 before miniaudio release a new version.

-----

Another solution: define `MA_NO_WASAPI` as workaround

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -227,6 +227,7 @@ jobs:
 
     - name: Patch miniaudio
       if: matrix.audiolib == 'miniaudio'
+      shell: bash
       run: |
         cd miniaudio
         patch < miniaudio.patch


purerosefallen:
DON'T use patch.
plz fork miniaudio and change repo path, just as how irrlicht did

edo9300:
You should define the `MA_COINIT_VALUE` macro and put the same value used by irrlicht

mercury233:
> You should define the `MA_COINIT_VALUE` macro and put the same value used by irrlicht

_Both [my irrlicht fork](https://github.com/search?q=repo%3Amercury233%2Firrlicht%20CoInit&type=code) and the original irrlicht 1.9 don't have `CIrrDeviceWin32WindowsVersionWMI.cpp` which is in your version._

And miniaudio is using [`COINIT_MULTITHREADED`](https://github.com/edo9300/irrlicht1-8-4/blob/1.9-custom/source/Irrlicht/CIrrDeviceWin32WindowsVersionWMI.cpp#L406)  as default.
```
#ifndef MA_COINIT_VALUE
#define MA_COINIT_VALUE    0   /* 0 = COINIT_MULTITHREADED */
#endif
```

edo9300:
> > You should define the `MA_COINIT_VALUE` macro and put the same value used by irrlicht
> 
> _Both [my irrlicht fork](https://github.com/search?q=repo%3Amercury233%2Firrlicht%20CoInit&type=code) and the original irrlicht 1.9 don't have `CIrrDeviceWin32WindowsVersionWMI.cpp` which is in your version._
> 
> And miniaudio is using [`COINIT_MULTITHREADED`](https://github.com/edo9300/irrlicht1-8-4/blob/1.9-custom/source/Irrlicht/CIrrDeviceWin32WindowsVersionWMI.cpp#L406) as default.
> 
> ```
> #ifndef MA_COINIT_VALUE
> #define MA_COINIT_VALUE    0   /* 0 = COINIT_MULTITHREADED */
> #endif
> ```

In my case the call actually setting up COM is OleInitialize, which sets it to apartment threaded, (in the file you pointed out com is de initialized after the function si executed). My guess is that the ime handling is internally initializing COM with that specific threading model beforehand.

purerosefallen:
https://github.com/Fluorohydride/ygopro/pull/2840
I believe this to be better. patch is strange all the time.

mercury233:
I don't think so. Fork repo may cause problem such as https://github.com/Fluorohydride/ygopro/pull/2815#issuecomment-2934223264

purerosefallen:
> I don't think so. Fork repo may cause problem such as [#2815 (comment)](https://github.com/Fluorohydride/ygopro/pull/2815#issuecomment-2934223264)

We can avoid by telling the related repo commit / PR in our PR description.

However, patch is even harder to maintain, and sometimes the patch may fail with no reasons.

PR #2834, opened 2025-06-10
use std::strncpy

salix5:
We should avoid strcpy.

@mercury233 
@purerosefallen


mercury233:
need tool function

salix5:
will change later

PR #2840, opened 2025-06-14
use patched miniaudio

mercury233:
alternate to https://github.com/Fluorohydride/ygopro/pull/2833

https://github.com/mercury233/miniaudio : Applied that patch to the codes in `miniaudio_split`, when the `dev` branch of upstream didn't.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -120,7 +120,7 @@ jobs:
     - name: Download miniaudio
       if: matrix.audiolib == 'miniaudio'
       run: |
-        git clone --depth=1 --branch 0.11.22 https://github.com/mackron/miniaudio
+        git clone --depth=1 https://github.com/mercury233/miniaudio
         cd miniaudio
         xcopy /Y extras\miniaudio_split\miniaudio.* .
         cd ..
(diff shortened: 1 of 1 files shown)


purerosefallen:
just for ref, in mycard:
- make `mycard/miniaudio` on mycard, master and develop branch
- Like how irrlicht was done https://code.moenext.com/mycard/ygopro/-/blob/develop/.ci/prepare-irrlicht.sh#L4 edit `prepare-miniaudio.sh`

mercury233:
_and another https://code.moenext.com/mycard/miniaudio-new with weird changes and not sync with any upstream?_

purerosefallen:
> _and another https://code.moenext.com/mycard/miniaudio-new with weird changes and not sync with any upstream?_

`***-new` exists because there was an outdated repo. Ages before there were `irrlicht`, `irrlicht-mac` so the new repo called `irrlicht-new`

Now it can be called `miniaudio` without `-new` because first one is good enough.

mercury233:
> Now it can be called `miniaudio` without `-new` because first one is good enough.

One day it will be not good enough if you don't sync with upstream.

purerosefallen:
> > Now it can be called `miniaudio` without `-new` because first one is good enough.
> 
> One day it will be not good enough if you don't sync with upstream.

syncing is done when a pre tag submitted, or during release.

irrlicht had some historic reasons, such as multiple versions, build method change. we believe the same thing would not occur in miniaudio if properly maintained

btw it's a bit OT to talk about mycard infra maintenance

purerosefallen:
could we use this one now?

mercury233:
> could we use this one now?

it is up to you

PR #2846, opened 2025-06-19
fix new replay crash

purerosefallen:


salix5:
Maybe we should use fseek

purerosefallen:
> Maybe we should use fseek

then you make one

purerosefallen:
https://chatgpt.com/share/6853cb2e-f788-800e-b8c8-09d554f01b18

some changes may required

PR #2847, opened 2025-06-19
read ExtendedReplayHeader

salix5:


salix5:
I think it would work but can be improved.

PR #2848, opened 2025-06-20
revert cppdialect

salix5:
Let the compiler decide.

@mercury233 
@purerosefallen 
@Wind2009-Louse
@fallenstardust

### DIFF
--- a/gframe/premake5.lua
+++ b/gframe/premake5.lua
@@ -3,7 +3,6 @@ include "spmemvfs/."
 
 project "YGOPro"
     kind "WindowedApp"
-    cppdialect "C++14"
     rtti "Off"
     openmp "On"
 
--- a/gframe/premake5.lua
+++ b/gframe/premake5.lua
@@ -88,7 +88,7 @@ project "YGOPro"
         end
 
     filter "system:macosx"
-        cppdialect "gnu++14"
+        cppdialect "C++14"
         openmp "Off"
         links { "OpenGL.framework", "Cocoa.framework", "IOKit.framework" }
         defines { "GL_SILENCE_DEPRECATION" }
--- a/gframe/premake5.lua
+++ b/gframe/premake5.lua
@@ -88,6 +88,7 @@ project "YGOPro"
         end
 
     filter "system:macosx"
+        cppdialect "C++14"
         openmp "Off"
         links { "OpenGL.framework", "Cocoa.framework", "IOKit.framework" }
         defines { "GL_SILENCE_DEPRECATION" }
--- a/gframe/premake5.lua
+++ b/gframe/premake5.lua
@@ -86,9 +86,10 @@ project "YGOPro"
                 filter {}
             end
         end
+    filter "not action:vs*"
+        cppdialect "C++14"
 
     filter "system:macosx"
-        cppdialect "C++14"
         openmp "Off"
         links { "OpenGL.framework", "Cocoa.framework", "IOKit.framework" }
         defines { "GL_SILENCE_DEPRECATION" }
--- a/gframe/premake5.lua
+++ b/gframe/premake5.lua
@@ -88,7 +88,7 @@ project "YGOPro"
         end
 
     filter "system:macosx"
-        cppdialect "C++14"
+        cppdialect "gnu++14"
         openmp "Off"
         links { "OpenGL.framework", "Cocoa.framework", "IOKit.framework" }
         defines { "GL_SILENCE_DEPRECATION" }
--- a/premake/irrlicht/premake5.lua
+++ b/premake/irrlicht/premake5.lua
@@ -162,7 +162,7 @@ project "irrlicht"
         end
 
     filter { "system:macosx" }
-        cppdialect "C++14"
+        cppdialect "gnu++14"
         defines { "GL_SILENCE_DEPRECATION" }
         undefines { "NO_IRR_COMPILE_WITH_JOYSTICK_EVENTS_" }
         files {
--- a/premake/irrlicht/premake5.lua
+++ b/premake/irrlicht/premake5.lua
@@ -162,7 +162,7 @@ project "irrlicht"
         end
 
     filter { "system:macosx" }
-        cppdialect "gnu++14"
+        cppdialect "C++14"
         defines { "GL_SILENCE_DEPRECATION" }
         undefines { "NO_IRR_COMPILE_WITH_JOYSTICK_EVENTS_" }
         files {


mercury233:
it seems that some compiler don't set c++14 as default yet

salix5:
Visual Studio and gcc have their default values.
Mac OS might have its own consideration.

purerosefallen:
I'm unwilling to see someone taking a random server and compile and breaks

salix5:
Irrlicht has stated that they have to be compiled under gnu++ in gcc.


> I'm unwilling to see someone taking a random server and compile and breaks

If the compiler says yes, everything is fine. 
if the compiler says no, one should ask the compiler, not ygopro.
ygopro is agonistic all along.

purerosefallen:
> Irrlicht has stated that they have to be compiled under gnu++ in gcc.
> 
> > I'm unwilling to see someone taking a random server and compile and breaks
> 
> If the compiler says yes, everything is fine. if the compiler says no, one should ask the compiler, not ygopro. ygopro is agonistic all along.

It's even worse if compiler says yes and then the server behaves weird or crash when running.

salix5:
if so, it is a bug in the source code, not a problem of CPP dialect.
it is quite confusing when one says going forward is wrong and another one says going backward is wrong.

mercury233:
How about we use highest avail standard in the debug mode, but use C++14 in the release mode?

purerosefallen:
> if so, it is a bug in the source code, not a problem of CPP dialect. it is quite confusing when one says going forward is wrong and another one says going backward is wrong.

The correct way to go forward is to state the C++ version. If going to C++ 17, then patch it to `cppdialect "c++17"`
But I don't think we should go *as far as possible* which is not the correct software engineering way.

> How about we use highest avail standard in the debug mode, but use C++14 in the release mode?

Then problems like *it runs on debug not on release*, or the opposite, would appear.

purerosefallen:
https://github.com/mycard/ygopro/pull/25

Note: that time the server failed to build then I added this line

mercury233:
> Then problems like _it runs on debug not on release_, or the opposite, would appear.

Then we fix them. Actually the purpose is to find those problems.

salix5:
> How about we use highest avail standard in the debug mode, but use C++14 in the release mode?

If we are going to move to c++17 or further:
Adding cppdialect might be useful for marking the current position.

If not:
Adding cppdialect is redundant and even harmful after several years. 
We should revert it and leave it to compiler.

I don't think we can take both of them.

salix5:
> The correct way to go forward is to state the C++ version. If going to C++ 17, then patch it to `cppdialect "c++17"`
> But I don't think we should go _as far as possible_ which is not the correct software engineering way.

Just in case that you did not notice: 
This PR is going backward.

mercury233:
> Just in case that you did not notice: This PR is going backward.

C++17 mode is the default since GCC 11, and ubuntu 24 has gcc 13

> If we are going to move to c++17 or further:
If not:
I don't think we can take both of them.

so you are taking both of them in this PR

salix5:
> so you are taking both of them in this PR

No.
cppdialect was added by me in #2730 , and now I think it is unnecessary and should be reverted back.
if you have any questions about the default mode in gcc,  please ask gcc, not me.

mercury233:
Right now, I feel that compiling separately with C++14 and C++17 as targets would be a choice that satisfies _some_ people.

_In #2730, you didn't specify a new C++ version; you only changed the way the version is specified from `-std=c++14` to `cppdialect`._

**If we migrate to C++17:**
This means that syntax removed in C++17 is no longer available, and new syntax introduced in C++17 is used. Therefore, compilation with C++17 is required.
Specifying the C++ version is currently necessary because MSVC, Apple Clang, and older versions of GCC do not default to C++17.

**If we stay with C++14:**
This means that syntax removed in C++17 is still available. Therefore, compilation with C++14 is required.
Specifying the C++ version is always necessary because GCC defaults to C++17 starting from version 11.

**If we compile separately with C++14 and C++17 as targets:**
This means that syntax removed in C++17 is no longer available, but new syntax introduced in C++17 also cannot be used.
Specifying the C++ version is still necessary because the default version in Apple Clang appears to be even lower than C++14.

salix5:
cppdialect 
reverted back to the state before #2730

mercury233:
> **If we stay with C++14:**
> This means that syntax removed in C++17 is still available. Therefore, compilation with C++14 is required.
> Specifying the C++ version is always necessary because GCC defaults to C++17 starting from version 11.

@salix5 so you have chosen this?

purerosefallen:
My opinion stays the same.
YGOPro should specify ONE C++ version, and apply to every target

salix5:
> > **If we stay with C++14:**
> > This means that syntax removed in C++17 is still available. Therefore, compilation with C++14 is required.
> > Specifying the C++ version is always necessary because GCC defaults to C++17 starting from version 11.
> 
> @salix5 so you have chosen this?

I choose to revert back to the previous state before I edited.
I don't know what it is, but I am sure the project did work at that time.

> My opinion stays the same. YGOPro should specify ONE C++ version, and apply to every target

The compiler will "specify ONE C++ version".

purerosefallen:
> The compiler will "specify ONE C++ version".

Different compilers specify different C++ versions, across vs2017, 2019, 2022, and so on, unless they use the same.

When it comes to gcc or Apple clang it's even worse.

We should make the build as reproducible as possible.

salix5:
If the compiler says yes, everything is fine.
You don't need to worry.

If the compiler says no, one should ask the compiler, not ygopro.
ygopro is agonistic all along.

purerosefallen:
> If the compiler says yes, everything is fine.

The program compiles well does not mean that the program runs well.

> If the compiler says no, one should ask the compiler, not ygopro.

Then they would report it as a bug, but in fact it's because the default compiler behavior differs.

salix5:
If the program compiles well but does not run well, it is a bug in the source code.
If you find it, you can fix it or open an issue.

If you firmly believe that cppdialect will cause some problems, can you give at least 1 example that is caused by cppdialect?

purerosefallen:
> If the program compiles well but does not run well, it is a bug in the source code. If you find it, you can fix it or open an issue.
> 
> If you firmly believe that cppdialect will cause some problems, can you give at least 1 example that is caused by cppdialect?

You may try to learn the spirit of NixOS, in which EVERY software versions are stable, and every gcc builds are reproducible, even with same sha256.
For server maintainance this is quite important, to make it be able to build on every OS, with steady results.

salix5:
In ygopro EVERY software versions are quite unstable, so I don't think there is anything to compare.
Can you give at least 1 example in ygopro that is caused by cppdialect?

PR #2849, opened 2025-06-21
remove CTOS_HS_READY

purerosefallen:
This packet does nothing, and `CTOS_UPDATE_DECK` can handle the signal of player ready.

In rare cases those packets may be in random order and cause players facing *0 main decks* error

note: old clients can join without problem, but their `CTOS_HS_READY` would be simply ignored

### DIFF
--- a/gframe/menu_handler.cpp
+++ b/gframe/menu_handler.cpp
@@ -184,7 +184,6 @@ bool MenuHandler::OnEvent(const irr::SEvent& event) {
 					break;
 				}
 				UpdateDeck();
-				DuelClient::SendPacketToServer(CTOS_HS_READY);
 				mainGame->cbCategorySelect->setEnabled(false);
 				mainGame->cbDeckSelect->setEnabled(false);
 				break;
@@ -636,7 +635,6 @@ bool MenuHandler::OnEvent(const irr::SEvent& event) {
 						break;
 					}
 					UpdateDeck();
-					DuelClient::SendPacketToServer(CTOS_HS_READY);
 					mainGame->cbCategorySelect->setEnabled(false);
 					mainGame->cbDeckSelect->setEnabled(false);
 				} else {
--- a/gframe/netserver.cpp
+++ b/gframe/netserver.cpp
@@ -337,11 +337,10 @@ void NetServer::HandleCTOSPacket(DuelPlayer* dp, unsigned char* data, int len) {
 		duel_mode->ToObserver(dp);
 		break;
 	}
-	case CTOS_HS_READY:
 	case CTOS_HS_NOTREADY: {
 		if (!duel_mode || duel_mode->pduel)
 			return;
-		duel_mode->PlayerReady(dp, (CTOS_HS_NOTREADY - pktType) != 0);
+		duel_mode->PlayerReady(dp, false);
 		break;
 	}
 	case CTOS_HS_KICK: {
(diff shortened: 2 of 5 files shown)


purerosefallen:
Side note: I have seen YGOPro2 clients sending CTOS_PLAYER_INFO and CTOS_JOIN_GAME in reverse order

mercury233:
Selecting a deck and clicking "Ready" in the game are not directly related. It might actually be necessary for the server to first respond with the result of the deck validation, and then the client proceed to send ctos_ready.

purerosefallen:
> Selecting a deck and clicking "Ready" in the game are not directly related. It might actually be necessary for the server to first respond with the result of the deck validation, and then the client proceed to send ctos_ready.

Then it's something else to do.
However, currently it's client sending two packets at once, and for some proxy situations, srvpro or the router thing, there may be async deck checks, in which the packet may arrive at server in different order.

For instance, MyCard entertain ranked match server (7911) would POST HTTP to check if the deck is valid to use, in which CTOS_UPDATE_DECK may arrive a bit slower, but CTOS_READY arrives in advance, causing errors. In addition, deck checking is also an async operation in srvpro deck_check

purerosefallen:
https://github.com/mycard/srvpro/blob/master/ygopro-server.coffee#L3315

This is an example that double CTOS for update_deck + ready leads to a problem.

With one single message for ready, we could simplify it to make a proper error message when invalid deck

purerosefallen:
note: planning to apply it to server before it merges
the `_setBuff` an invalid deck when deck check failing seemed to be too stupid

purerosefallen:
Can we apply this?
This is tested in Chocolate, and the latest SRVPro requires this.

mercury233:
> Can we apply this? This is tested in Chocolate, and the latest SRVPro requires this.

The other developers won't have time to deal with this.

purerosefallen:
planning on 0x1363?

purerosefallen:
@salix5 
have a look?

juice928:
👋 Hi, I'm an automated AI code review bot. I ran some checks on this PR and found **2 points that might be worth attention** (could be false positives, please use your judgment):

1. <strong>The player is automatically marked as 'Ready' even if deck validation fails in Single Duel mode</strong>
   - [Location](https://github.com/Fluorohydride/ygopro/blob/d9d0cb8c0977f7c18b78a690a926066796728dea/gframe/single_duel.cpp#L306): `gframe/single_duel.cpp:L306`
   - Impact: This could allow players to enter matches with illegal or banned decks, bypassing the intended validation logic.
   - Suggestion: Consider wrapping the `PlayerReady` call in a conditional check to ensure it only triggers when `deck_error[dp->type]` is clear.

2. <strong>The 'Ready' state is triggered unconditionally during deck updates in Tag Duel mode</strong>
   - [Location](https://github.com/Fluorohydride/ygopro/blob/d9d0cb8c0977f7c18b78a690a926066796728dea/gframe/tag_duel.cpp#L286): `gframe/tag_duel.cpp:L286`
   - Impact: Similar to the Single Duel logic, players might be forced into a ready state despite having errors in their loaded decks.
   - Suggestion: It would be more robust to verify that `deck_error` is zero before confirming the player is ready for the duel.

If you find these suggestions disruptive, you can reply "stop" , and I'll automatically skip this repository in the future.

PR #2854, opened 2025-06-23
remove unnecessary features from Freetype

mercury233:
YGOPro only uses TrueType and OpenType fonts, so other Freetype features can be removed.

After removal, the Windows 64-bit exe size is reduced by about 360KB.


purerosefallen:
Does it mean which of the following:
- Our distro packs TrueType fonts
- YGOPro (without this pr) does not read any non-TrueType fonts

purerosefallen:
It works on Linux with MyCard packed fonts.
But other fonts untested.

PR #2856, opened 2025-06-24
fix animation of card moving to extra

mercury233:
**Before (10x slower):**
before

**Fixed:**
after

**Problem:** 
The `code` of the card, which is moving to extra, is reset to 0 before the animation complete, in 
https://github.com/Fluorohydride/ygopro/blame/ed09b7a36f03a27163429cb324098544d481ecb1/gframe/duelclient.cpp#L2620
(I doubt whether it is necessary to specify `0x40(LOCATION_EXTRA)`)
or in 
https://github.com/Fluorohydride/ygopro/blame/ed09b7a36f03a27163429cb324098544d481ecb1/gframe/single_duel.cpp#L967

**Solution:**
Use a temp code in animation. Assume that if a card had code before but is set to 0 now, it must had been moved.
`chain_code` is used before in `ShowSelectCard`(`conti_selecting`), and we can use this. Maybe we should change the variable name.

### DIFF
--- a/gframe/client_card.cpp
+++ b/gframe/client_card.cpp
@@ -35,8 +35,12 @@ void ClientCard::SetCode(unsigned int x) {
 	if((location == LOCATION_HAND) && (code != x)) {
 		code = x;
 		mainGame->dField.MoveCard(this, 5);
-	} else
+	} else {
+		if (x == 0 && code != 0) {
+			chain_code = code;
+		}
 		code = x;
+	}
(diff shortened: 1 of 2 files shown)


purerosefallen:
How about card moving to deck, or moving to banish face down?

mercury233:
> How about card moving to deck, or moving to banish face down?

also fixed by this changement

PR #2858, opened 2025-06-27
remove -fno-strict-aliasing

salix5:


mercury233:
I'm not very familiar with C++. Are you sure we won't run into strict aliasing issues when handling `return_value returns`?

mercury233:
> The fact is, Linux uses -fno-strict-aliasing for a damn good reason: 
because the gcc notion of "strict aliasing" is one huge stinking pile of 
sh*t. Linux doesn't use that flag because Linux is playing fast and loose, 
it uses that flag because _not_ using that flag is insane.
https://www.mail-archive.com/linux-btrfs@vger.kernel.org/msg01647.html

Linus Torvalds was strongly against strict aliasing.

PR #2862, opened 2025-06-29
add ReplayMode::LoadReplay

salix5:
# Problem
圖片
```
ygopro -r asdfzxcvbnm.yrp
```
If ygopro is executed in command line with a wrong filename, the error will occur.

# Reason
https://github.com/Fluorohydride/ygopro/blob/f0e9bc883176cc66969d0321c8aa30bb6d783456/gframe/menu_handler.cpp#L237

It does not check the return value of OpenReplay.
Moreover, sending the click event manually is not a good way.

# Solution
BUTTON_LOAD_REPLAY and command line argument will call this function.

@mercury233 
@purerosefallen

### DIFF
--- a/gframe/gframe.cpp
+++ b/gframe/gframe.cpp
@@ -1,6 +1,7 @@
 #include "config.h"
 #include "game.h"
 #include "data_manager.h"
+#include "replay_mode.h"
 #include <event2/thread.h>
 #include <clocale>
 #include <memory>
@@ -11,10 +12,10 @@
 unsigned int enable_log = 0x3;
 bool exit_on_return = false;
 bool open_file = false;
-wchar_t open_file_name[256] = L"";
+wchar_t open_file_name[256]{};
 bool bot_mode = false;
 
-void ClickButton(irr::gui::IGUIElement* btn) {
+static void ClickButton(irr::gui::IGUIElement* btn) {
 	irr::SEvent event;
 	event.EventType = irr::EET_GUI_EVENT;
 	event.GUIEvent.EventType = irr::gui::EGET_BUTTON_CLICKED;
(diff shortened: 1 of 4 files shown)


gframe/gframe.cpp
```diff
@@ -100,7 +101,7 @@ int main(int argc, char* argv[]) {
 				BufferIO::CopyWideString(wargv[1], open_file_name);
 				exit_on_return = true;
 				ClickButton(ygo::mainGame->btnReplayMode);
```
purerosefallen, line 103: Is this still required?

gframe/replay_mode.cpp
```diff
@@ -20,10 +20,34 @@ int ReplayMode::skip_turn = 0;
...
 
+void ReplayMode::LoadReplay(const wchar_t* replay_path, int start_turn) {
+	if (!ReplayMode::cur_replay.OpenReplay(replay_path)) {
+		if (exit_on_return)
+			mainGame->device->closeDevice();
+		return;
+	}
+	open_file = false;
+	mainGame->ClearCardInfo();
+	mainGame->wCardImg->setVisible(true);
```
mercury233, line 31: I think increasing coupling is not a good idea.

salix5: 什麼意思 看不太懂

gframe/gframe.cpp
salix5: I think yes.
LoadReplay starts from loading replay file.
In normal route, it happens after clicking BUTTON_LOAD_REPLAY.

gframe/replay_mode.cpp
mercury233: replay_mode.cpp 应该专注于处理录像，涉及图形界面的东西越少越好

salix5: 如果你的意思是增加各個模組間的依賴程度
我想加入新函數應該是降低依賴程度才對

```
ygopro -r 123.yrp
```
原本：
手動觸發EGET_BUTTON_CLICKED事件
進入
gframe/menu_handler.cpp
BUTTON_LOAD_REPLAY

現在：
ReplayMode::LoadReplay

salix5: > replay_mode.cpp 应该专注于处理录像，涉及图形界面的东西越少越好

那移到MenuHandler?

PR #2879, opened 2025-07-13
add LUA_USE_LONGJMP

purerosefallen:
ygopro builds Lua in C++ but without a proper catch for C++ exceptions. This sometimes crashes YGOPro, especially in dynamic linking situations.

### DIFF
--- a/premake/lua/premake5.lua
+++ b/premake/lua/premake5.lua
@@ -5,6 +5,8 @@ project "lua"
     files { "src/*.c", "src/*.h" }
     removefiles { "src/lua.c", "src/luac.c", "src/linit.c", "src/onelua.c" }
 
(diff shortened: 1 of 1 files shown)


edo9300:
C++ exceptions are required in lua due since without that c++ objects allocated on the stack become corrupt when lua yields or an error is thrown.
As for the crashes, it seems a weird thing to me, i had encountered a tangentially related crash while building a static binary with musl with lto enabled, likely a side effect of the extreme optimization.
You can see the stack unwinding issue by yourself with this piece of code
https://github.com/edo9300/ygopro-core/blob/e2f5d6de3fbd85d6c126e9adbb32c9e15c34c50b/interpreter.cpp#L39-L58

PR #2885, opened 2025-07-22
add `mysnprintf`, use `std::forward` in `myswprintf`

mercury233:
A simple wrapper around `std::snprintf` to simplify its usage.

- We already have `myswprintf` that does the same thing.
- `ocgcore` also has `interpreter::sprintf` that does the same thing.

### DIFF
--- a/gframe/config.h
+++ b/gframe/config.h
@@ -54,12 +54,12 @@
 #include "../ocgcore/ocgapi.h"
 
 template<size_t N, typename... TR>
-inline int myswprintf(wchar_t(&buf)[N], const wchar_t* fmt, TR... args) {
-	return std::swprintf(buf, N, fmt, args...);
+inline int myswprintf(wchar_t(&buf)[N], const wchar_t* fmt, TR&&... args) {
+	return std::swprintf(buf, N, fmt, std::forward<TR>(args)...);
 }
 template<size_t N, typename... TR>
-inline int mysnprintf(char(&buf)[N], const char* fmt, TR... args) {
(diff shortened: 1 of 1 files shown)


salix5:
I think the copilot is right.
char is 1 byte, and sizeof(soundPath) is enough.

mercury233:
> I think the copilot is right. char is 1 byte, and sizeof(soundPath) is enough.

It was wrong. The code is equivalent.

salix5:
My comment is not clear.

What I mean is:
Adding a new function for calculating array length is not quite necessary.
```cpp
std::snprintf(buf, sizeof buf, "%d", 123);

// or
std::snprintf(buf, std::size(buf), "%d", 123);
```
is more intuitive for both human and AI.

Wind2009-Louse (approved):

salix5:
snprintf沒有必要用另外一個函數包裝
我能理解的只有“你不喜歡sizeof”
這很難說是合理的理由

mercury233:
从 https://github.com/Fluorohydride/ygopro/pull/2540 就最好添加对 snprintf 的包装以保持一致性

salix5:
#2540 
myswprintf 
都只是以前因為時代背景遺留的各種問題
現代無法大改只能勉強跟已有的函數妥協
那也沒有辦法

為了"一致性"往這堆上面再加新的函數
我只能理解為你實在無法忍受sizeof 
而且snprintf 的參數都是primitive type 
原本就不需要std::forward

PR #2887, opened 2025-07-26
Remove WinXP support from build configuration

salix5:
@mercury233 
@purerosefallen 
@Wind2009-Louse
@fallenstardust

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -83,7 +83,6 @@ newoption { trigger = "irrklang-pro-release-lib-dir", category = "YGOPro - irrkl
 newoption { trigger = "irrklang-pro-debug-lib-dir", category = "YGOPro - irrklang - pro", description = "", value = "PATH" }
 newoption { trigger = 'build-ikpmp3', category = "YGOPro - irrklang - ikpmp3", description = "" }
 
-newoption { trigger = "winxp-support", category = "YGOPro", description = "" }
 newoption { trigger = "mac-arm", category = "YGOPro", description = "Compile for Apple Silicon Mac" }
 newoption { trigger = "mac-intel", category = "YGOPro", description = "Compile for Intel Mac" }
 
@@ -233,10 +232,6 @@ if USE_AUDIO then
     end
 end
 
-if GetParam("winxp-support") and os.istarget("windows") then
-    WINXP_SUPPORT = true
-end
-
 if os.istarget("macosx") then
     if GetParam("mac-arm") then
         MAC_ARM = true
@@ -260,12 +255,7 @@ workspace "YGOPro"
     filter "system:windows"
         systemversion "latest"
         startproject "YGOPro"
-        if WINXP_SUPPORT then
-            defines { "WINVER=0x0501" }
-            toolset "v141_xp"
-        else
-            defines { "WINVER=0x0601" } -- WIN7
-        end
+        defines { "WINVER=0x0601" } -- WIN7
         platforms { "Win32", "x64" }
 
     filter { "system:windows", "platforms:Win32" }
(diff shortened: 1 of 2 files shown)


mercury233:
One more important thing is to remove `FOPEN_WINDOWS_SUPPORT_UTF8` and create a compile error if the Windows SDK is below the required version.

mercury233:
Should we remove `myfopen`? I prefer to keep it and add `mysnprintf`.

salix5:
> Should we remove `myfopen`? I prefer to keep it and add `mysnprintf`.

myfopen and mysnprintf are different.

myfopen
- Solves real cross-platform issues
Handles UTF-8 filename encoding automatically on Windows vs Unix

- Provides actual functionality
Without it, we would need `#ifdef` blocks everywhere for file operations

- Good abstraction
Hides platform complexity behind a clean interface

myswprintf 
- Pure syntactic sugar
Only saves typing one parameter (the buffer size)

- No functional value
std::swprintf(buf, std::size(buf), fmt, args...) does exactly the same thing in C++17

- Adds cognitive overhead
Developers need to understand the template magic instead of using standard library

----
以前加入myswprintf可能是在當時的環境有必要
在現代這樣做的幫助已經非常有限

## 陣列長度
```c
wchar_t buf[256];
```

```
swprintf(buf, 256, L"%d", 123);
```
這個方法的缺陷很明顯：
256是肉眼看到然後手動輸入的數值
人類看錯或以後buf的長度改變都可能產生錯誤
而且這是可以自動計算的數值
原本就不需要手動輸入

```c
swprintf(buf, sizeof buf / sizeof buf[0], L"%d", 123);
```
自動計算陣列長度
C標準建議的寫法
可能的缺點：字數很多輸入費時？


```cpp
template<size_t N, typename... TR>
inline int myswprintf(wchar_t(&buf)[N], const wchar_t* fmt, TR... args) {
	return std::swprintf(buf, N, fmt, args...);
}

myswprintf(buf, L"%d", 123);

// C++17
std::swprintf(buf, std::size(buf), L"%d", 123);
```
myswprintf
優點：字數較少輸入省時
缺點：只為轉發標準函數的自訂函數會降低可讀性

就連這個優點在現代來看也幫助不大
C++17加入std::size計算長度
現代已經不需要用這種迂迴的方法計算陣列長度
因此就連myswprintf是否仍有必要都有疑問

```cpp
char buf[256];
std::snprintf(buf, sizeof buf, "%d", 123);

// C++17
std::snprintf(buf, std::size(buf), "%d", 123);
```
snprintf
由於char的長度=1
sizeof取得的大小就是陣列長度
因此原本就沒有特地用其他函數轉發的需求
在C++17加入std::size以後更是如此

我認為以現代的角度來看
myswprintf是否仍有必要保留都有疑問
似乎沒有必要再另外加入mysnprintf

mercury233:
I mean, after https://github.com/Fluorohydride/ygopro/pull/2889 , `myfopen` may be no longer needed.

-----

I don't think `myswprintf` reduces readability; on the contrary, I think it actually improves it. In fact, one thing I personally dislike about “modern C++” is that the massive amount of special syntax and keywords significantly reduces readability.

premake5.lua
```diff
@@ -233,10 +232,6 @@ if USE_AUDIO then
     end
 end
 
-if GetParam("winxp-support") and os.istarget("windows") then
```
mercury233, line 236: better add warning instead

salix5:
Why would the function myswprintf, which is only used by ygopro, improve readibility?
Everyone knows swprintf, sizeof, std::size, but I guess very few people has heard of `myswprintf`.

Wind2009-Louse (approved):

PR #2888, opened 2025-07-27
turn off TRUSTED_SCHEMA for security

salix5:
https://www.sqlite.org/security.html
If the application includes any custom SQL functions or custom virtual tables that have side effects or that might leak privileged information, then the application should use one or more of the techniques below to prevent a maliciously crafted database schema from surreptitiously running those SQL functions and/or virtual tables for nefarious purposes:
- Invoke sqlite3_db_config(db,SQLITE_DBCONFIG_TRUSTED_SCHEMA,0,0) on each database connection as soon as it is opened.


mercury233:
I don't think it's necessary, because the program only reads from each database once. At most, a maliciously crafted database could only cause the program to crash. If there were more serious consequences, enabling this feature might not be able to prevent them either.

PR #2891, opened 2025-07-31
change card_data::level to 2 bytes

salix5:
# The disadvantage of packing
Packing multiple unrelated properties into the same column is considered a poor practice in database design.

- Violates database normalization
Normalization is a design principle that separates data into logical tables and columns to reduce redundancy and improve clarity.
Packing multiple attributes into one column breaks this rule, and it makes the schema less readable, harder to understand, and more error-prone.

- Harder to query
Queries are harder to write and maintain.
You must remember which bit means what.
Even simple queries become more complex, requiring constant use of bitwise operations.

- Poor indexing support
SQLite indexes individual columns, but not individual bits inside an integer.
If you pack 4 values  into one INTEGER column, it cannot index each flag efficiently.
Queries on individual bits won’t benefit from indexes, resulting in full table scans.
It can hurt performance, especially when the table grows.

- Difficult to maintain
With individual columns, adding or changing values is straightforward.
But when values are packed into a single field, even basic read or write operations require bitwise manipulation.
For example: 
What is the level of a card with a `level` column of 50528266?
If the level needs to be changed to 11, what is the new value of `level`?

- Minimal space savings in SQLite
SQLite is already very efficient at storing integers.
It uses variable-length integers internally, which take only as many bytes as needed.
The space saving from packing bits is negligible.

# What can ygopro do now?
It is too late to change the behavior of packing the pendulum scale into `level` column.
But ygopro can do something to prevent it from becoming worse.

What this PR tries to do:
- The cdb file remains unchanged.
- If new card properties are introduced in the future, they should be stored in new columns.
- No additional values will be packed into the existing `level` column.
- The lower 16 bits of the level column are now reserved exclusively for storing the level/rank/link value.

----

level column in cdb (from most significant bit)
Before:
first byte: lscale
second byte: rscale
third byte: (not used)
fourth byte: level

After:
first byte: lscale
second byte: rscale
3-4 byte: level

@mercury233 
@purerosefallen 
@Wind2009-Louse
@fallenstardust

### DIFF
--- a/gframe/data_manager.cpp
+++ b/gframe/data_manager.cpp
@@ -38,7 +38,7 @@ bool DataManager::ReadDB(sqlite3* pDB) {
 		else
 			cd.link_marker = 0;
 		uint32_t level = static_cast<uint32_t>(sqlite3_column_int64(pStmt, 7));
-		cd.level = level & 0xff;
+		cd.level = level & 0xffff;
 		cd.lscale = (level >> 24) & 0xff;
 		cd.rscale = (level >> 16) & 0xff;
 		cd.race = static_cast<decltype(cd.race)>(sqlite3_column_int64(pStmt, 8));


mercury233:
why?

salix5:
Why should ygopro use only 1 byte when there are 2?

mercury233:
There are 8 and we only use 3 bytes.

salix5:
This column is uint32 from at least 10 years ago.
Do you want to change it to uint64?

If not, then there are only 4 bytes.
Why should ygopro skip one byte?

mercury233:
https://github.com/Fluorohydride/ygopro/commit/a7eb57f5a2b125525061ea6d2e56a19fd8fcd5ec#diff-9facdbf97ca4496f5c254007a7c22f3e6878af80aef2c118f51db16866b4d180R35

The "level" field has been defined as 1 byte for over 10 years too, and there is neither sufficiency nor necessity to modify it.

salix5:
Sqlite database is designed for one value per column. 
Packing 3 different value into one column by bitwise operation is not appropriate in the first place.
ygopro should have added a new column to store pendulum scale.

Even if it has to pack 3 values into 1 column for some unknown reason,  you still didn't answer the question:
Why should ygopro skip one byte in uint32?

Edit:
I hope your answer is not something like 
"Because it is the law of ancestors." 
not very convincing to me.

purerosefallen:
Disagree.
NO ONE in ygo community, including DIY or official, make level more than 255.

if there's a new field in future yugioh ruling, we could use that byte that time. However it's not wise to occupy that byte today.

salix5:
It sounds like you are preparing to pack the fourth value into the same column. 
That is quite astonishing.

purerosefallen:
> It sounds like you are preparing to pack the fourth value into the same column. That is quite astonishing.

Occupying the remaining byte to a useless feature is even more astonishing.
Or could you tell any situations where level >255 would be involved?

salix5:
The level can be any int32 value.
If ygopro has to occupy 2 bytes of this column, at least it can be any uint16 value.

I think using int32 value is quite normal.
Packing 3,4 or 5 values into the same integer is surprising.

purerosefallen:
> The level can be any int32 value.

I even think the level field only needs 4 bits, for up to level 15.
Beyond that is used for abnormal cards, which should not be supported in ygopro.

salix5:
If some new cards whose original level is 500 are added in the 15th series, should ygopro accept these cards?
Or should ygopro reject these cards because you don't like them?

mercury233:
The issue is that we need a reason to make changes. You should provide a justification for the _modification_, while others are not obligated to explain why it should _remain unchanged_.

salix5:
# The disadvantage of packing
Packing multiple unrelated properties into the same column is considered a poor practice in database design.

- Violates database normalization
Normalization is a design principle that separates data into logical tables and columns to reduce redundancy and improve clarity.
Packing multiple attributes into one column breaks this rule, and it makes the schema less readable, harder to understand, and more error-prone.

- Harder to query
Queries are harder to write and maintain.
You must remember which bit means what.
Even simple queries become more complex, requiring constant use of bitwise operations.

- Poor indexing support
SQLite indexes individual columns, but not individual bits inside an integer.
If you pack 4 values  into one INTEGER column, it cannot index each flag efficiently.
Queries on individual bits won’t benefit from indexes, resulting in full table scans.
It can hurt performance, especially when the table grows.

- Difficult to maintain
With individual columns, adding or changing values is straightforward.
But when values are packed into a single field, even basic read or write operations require bitwise manipulation.
For example: 
What is the level of a card with a `level` column of 50528266?
If the level needs to be changed to 11, what is the new value of `level`?

- Minimal space savings in SQLite
SQLite is already very efficient at storing integers.
It uses variable-length integers internally, which take only as many bytes as needed.
The space saving from packing bits is negligible.


# What can ygopro do now?
It is too late to change the behavior of packing the pendulum scale into `level` column.
But ygopro can do something to prevent it from becoming worse.

What this PR tries to do:
- The cdb file remains unchanged.
- If new card properties are introduced in the future, they should be stored in new columns.
- No additional values will be packed into the existing `level` column.
- The lower 16 bits of the level column are now reserved exclusively for storing the level/rank/link value.

salix5:
# 打包的缺點
在資料庫設計中，將多個不相關的屬性打包到同一欄位被認為是一種糟糕的做法。

- 違反資料庫規範化
規範化是一種設計原則，它將資料分成邏輯表和欄位，以減少冗餘並提高清晰度。
將多個屬性打包到一列違反了此規則，會使模式的可讀性降低、更難理解，並且更容易出錯。

- 查詢難度加大
查詢的編寫和維護難度加大。
你必須記住每個位元代表什麼。
即使是簡單的查詢也會變得更加複雜，需要不斷使用位元運算。

- 索引支援不佳
SQLite 會索引單一列，但不會索引整數中的單一位元。
如果將 4 個值打包到一個 INTEGER 欄位中，它無法有效地索引每個標誌。
針對單一位元的查詢不會從索引中受益，會導致全表掃描。
這會降低效能，尤其是在表規模成長時。

- 難以維護
使用單獨的欄位，新增或變更值非常簡單。
但是，當值被打包到單一欄位中時，即使是基本的讀寫操作也需要按位操作。
例如：
level 為 50528266 的卡牌的等級是多少？
如果需要將等級改為 11，那麼level的新值是多少？

- SQLite 中的最小空間節省
SQLite 在儲存整數方面已經非常有效率。
它內部使用可變長度整數，這些整數只佔用所需的位元組數。
按位打包節省的空間微乎其微。

# ygopro 現在能做什麼？
現在更改將靈擺刻度打包到level的行為已經太晚了。
但 ygopro 可以採取一些措施來防止情況惡化。

這個PR的目的：
- cdb 檔案保持不變。
- 如果將來引入新的卡牌屬性，則將其儲存在新的欄位。
- 現有的level欄位中不會再填入任何新的數值。
- level欄位的最低 16 位元現在專門用於儲存等級/等級/連結值。

purerosefallen:
we don't have plan to change cdb format, which requires cross-project changes.
eg. windbot, srvpro, ygopro-arena-api, ygopro-analytics, ...

If you wish to change cdb format plz open PR in every related projects you know.

salix5:
> we don't have plan to change cdb format, which requires cross-project changes.
> eg. windbot, srvpro, ygopro-arena-api, ygopro-analytics, ...

Does that "we" represent for the opinions of all projects in your list:
windbot, srvpro, ygopro-arena-api, ygopro-analytics, ...

purerosefallen:
> > we don't have plan to change cdb format, which requires cross-project changes.
> > eg. windbot, srvpro, ygopro-arena-api, ygopro-analytics, ...
> 
> Does that "we" represent for the opinions of all projects in your list: windbot, srvpro, ygopro-arena-api, ygopro-analytics, ...

that's the whole Chinese ygopro ecosystem, and those devs may not have time to catchup the change.

salix5:
> > > we don't have plan to change cdb format, which requires cross-project changes.
> > > eg. windbot, srvpro, ygopro-arena-api, ygopro-analytics, ...
> > 
> > 
> > Does that "we" represent for the opinions of all projects in your list: windbot, srvpro, ygopro-arena-api, ygopro-analytics, ...
> 
> that's the whole Chinese ygopro ecosystem, and those devs may not have time to catchup the change.

Does that "we" represent for the opinions of
> the whole Chinese ygopro ecosystem

purerosefallen:
> > > > we don't have plan to change cdb format, which requires cross-project changes.
> > > > eg. windbot, srvpro, ygopro-arena-api, ygopro-analytics, ...
> > > 
> > > 
> > > Does that "we" represent for the opinions of all projects in your list: windbot, srvpro, ygopro-arena-api, ygopro-analytics, ...
> > 
> > 
> > that's the whole Chinese ygopro ecosystem, and those devs may not have time to catchup the change.
> 
> Does that "we" represent for the opinions of
> 
> > the whole Chinese ygopro ecosystem

at least me and mercury would consider the whole ecosystem before making any breaking changes, and make sure the related devs have time to do so.
otherwise it would block the release progress

mercury233:
@purerosefallen Please be cautious when using "we“. I've had to change it to "I" several times already.

mercury233:
Indeed, there are some issues with the current database design, for example, the `setcode` field only allows 4 setcodes, the `category` classification doesn't accommodate effect development, and link arrows are being stored in the `def` field.

I think we may need an opportunity or proper occasion to optimize it, rather than making sudden changes.

_Even if, in the future, Xyz/Link monsters are officially given Levels on their card texts, I would still recommend storing that information in the second byte of the `level` field._

PR #2900, opened 2025-08-11
use cp -pr to copy out premake

purerosefallen:
yet another alt to https://github.com/Fluorohydride/ygopro/pull/2898

but stupid enough because most CIs, as well as dev envs for everyone, should be replaced

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -437,8 +437,8 @@ jobs:
 
     - name: Copy premake files
       run: |
-        cp -r premake/* .
-        cp -r resource/* .
+        cp -pr premake/* .
+        cp -pr resource/* .
 
     - name: Use premake to generate make files (apt packages)
       if: matrix.static-link != true
@@ -664,8 +664,8 @@ jobs:
 
     - name: Copy premake files
       run: |
-        cp -r premake/* .
-        cp -r resource/* .
+        cp -pr premake/* .
+        cp -pr resource/* .
 
     - name: Use premake to generate make files (Homebrew packages)
       if: matrix.static-link != true


mercury233:
**问题：**
在每次编译前，有些人可能会重复执行 premake 步骤，即重新将 `premake` 文件夹中的文件复制到各个目录，并再次运行 `premake5`。然而，复制代码文件可能会导致目标文件的修改时间被更新为当前时间，从而触发编译器对相关文件进行不必要的重新编译。具体来说，在 #2854 中新增的 `premake/freetype/custom` 下的三个头文件就会引发这一问题。

**解决方法：**
实际上，每次编译前都重新执行 premake 并不是必要的，大多数情况下也没有人会这么做。如果确实有这种需求，只需要在使用 `cp` 命令时加上 `-p` 选项以保留文件的时间戳。
在 Windows 下，无论是使用资源管理器复制还是 `xcopy`，默认都会保留文件的修改时间，因此一般不会导致这个问题。

---

**Problem:**
Before each build, some people may choose to re-run the premake step, which involves copying files from the `premake` folder into various directories and then running `premake5` again. However, copying code files can update the modification timestamps of the target files to the current time, which in turn causes the compiler to unnecessarily recompile affected files. Specifically, the three header files added in #2854 under `premake/freetype/custom` can trigger this issue.

**Solution:**
Re-running premake before every build is unnecessary, and in practice, very few people actually do this. If such a workflow is indeed required, simply add the `-p` option to the `cp` command to preserve file timestamps.
On Windows, both File Explorer copy operations and `xcopy` preserve modification times by default, so this problem does not occur.

PR #2903, opened 2025-08-12
add datas2 table to cdb & add setcode1-3

purerosefallen:
```sql
CREATE TABLE datas2(id integer primary key,setcode1 integer,setcode2 integer,setcode3 integer)
```

it works cross-version though the old client cannot read the extra setcodes.

require https://github.com/Fluorohydride/ygopro-core/pull/785

### DIFF
--- a/gframe/data_manager.cpp
+++ b/gframe/data_manager.cpp
@@ -15,53 +15,96 @@ DataManager::DataManager() : _datas(32768), _strings(32768) {
 		{55088578u, {0x8f, 0x54, 0x59, 0x82, 0x13a}},
 	};
 }
+static bool table_exists(sqlite3* db, const char* name) {
+	static const char* q =
+		"SELECT 1 FROM sqlite_master WHERE type='table' AND name=?1";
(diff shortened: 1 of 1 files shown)


mercury233:
not a good idea

salix5:
table_exists is unnecessary 
sqlite api already has metadata function 
```c
bool has_datas2 = (SQLITE_OK == sqlite3_table_column_metadata(pDB, nullptr, "datas2", "setcode1", nullptr, nullptr, nullptr, nullptr, nullptr))
```

purerosefallen:
> table_exists is unnecessary sqlite api already has metadata function
> 
> ```c
> bool has_datas2 = (SQLITE_OK == sqlite3_table_column_metadata(pDB, nullptr, "datas2", "setcode1", nullptr, nullptr, nullptr, nullptr, nullptr))
> ```

then it's better to change the impl of `table_exists`.

Note that there may be more tables in the future.

purerosefallen:
may work more on https://github.com/Fluorohydride/ygopro/pull/3044

PR #2919, opened 2025-09-03
Update .gitignore

salix5:


mercury233:
The developers may create symbol links for pics and deck too. There is no reason to change those in .gitignore, since `/abc` can ignore folder too.

PR #2929, opened 2025-09-22
2025.10 OCG & 2025.9 TCG lflist

jwyxym:


purerosefallen:
plz make TCG banlist at the 2nd place

PR #2931, opened 2025-09-27
add credit system in lflist.conf & support Genesys lflist

purerosefallen:
```
!2025.09 TCG Genesys

$genesys 100

# Genesys points
21044178 $genesys 100 -- Abyss Dweller
9464441 $genesys 20 -- Adamancipator Risen - Dragite
62320425 $genesys 50 -- Agido the Ancient Sentinel
38811586 $genesys 33 -- Albion the Sanctifire Dragon
1475311 $genesys 5 -- Allure of Darkness
```

Multiple credit namespace also supported

related projects:
- https://code.moenext.com/mycard/ygopro-rush-duel/-/blob/master/lflist.conf
- https://code.moenext.com/mycard/genesys-gen


mercury233:
It is not a good idea to use static picture for the mark. Shouldn't assume the credit of every card is below 100.

purerosefallen:
> It is not a good idea to use static picture for the mark. Shouldn't assume the credit of every card is below 100.

design comes from the online editor thing.
any better ways?

Or we could think of a way when a new rule comes out with single card credit >100

mercury233 (changes requested):
Shouldn't assume the credit of every card is below 100.

purerosefallen:
> Shouldn't assume the credit of every card is below 100.

When it comes to >100 credit, the deck check in deck editor / in-game would work as usual, but the graphic number label would not display
I would appreciate for another plan for displaying >100 numbers & non-static texture display

mercury233:
credit -> point
> Deck construction uses a **point** system. Some cards are assigned a **point** value;
https://www.yugioh-card.com/en/genesys/

salix5:
以前的lflist.conf都是預設從每行讀取2個整數
我是建議genesys分數可以用相同的格式
開頭用其他符號區分就好
例如

```
$2025.09 TCG Genesys
21044178 100 -- Abyss Dweller
```

另外官方網站有提到
4. The standard point cap is 100, but events can be run with any point cap, or even a zero-point cap! Official Tournament Stores can set their own caps for their tournaments. 

官方的分數最高看起來就是100
再加入標示>100的圖案應該就行

purerosefallen:
> 以前的lflist.conf都是預設從每行讀取2個整數 我是建議genesys分數可以用相同的格式 開頭用其他符號區分就好 例如

## Problems on the suggested format

Sometimes an ordinary banlist (0/1/2) AND a credit system should exist on the same banlist.

https://cdntx.moecube.com/ygopro-genesys/lflist.conf

Genesys is also an example that all Link & Pendulums are banned, in which `XXXXX 0` would exist.

## Design of this PR

This system is not only for Genesys, but also for other specific deck build limits, so I added a design for allowing multiple point namespaces.

Below Rush Duel is an example, for Legend card.

https://code.moenext.com/mycard/ygopro-rush-duel/-/blob/master/lflist.conf

Also there are other lflist in communities, something like DL-like server and so-on.

Btw I just announced this lflist format to them, so no plan to change. Otherwise there would be extra communication costs.

purerosefallen:
> credit -> point
> 
> > Deck construction uses a **point** system. Some cards are assigned a **point** value;
> > https://www.yugioh-card.com/en/genesys/

I'm not that willing to change the noun using in code, as long as there are no slight grammar mistakes. Btw there's a fork released just this update, with the filename in texture file.

In addition, the system design is not only for Genesys. So I don't think we should take the exact word from Genesys.

btw in YGOPro we say `lflist` while the official name is `banlist` or `Forbidden & Limit list`. Why wouldn't we change that in every piece of code?

I believe every contributor are able to choose the variable name in case there are no slight grammar problems.

purerosefallen:
> 再加入標示>100的圖案應該就行

I'm not a fan of over-engineering

purerosefallen:
In addition, I'm not planning to implement everything in Genesys, but a system with credit-system banlist. I would NOT do something like point total value display / Pendulum & Link disable / separate gamemode etc.

I'm just creating an interface that a Genesys server / environment can be built upon this, as well as other credit-like environments like Legend in Rush Duel & one limited / semi in DL. I believe they're community works, and let's leave it to the community.

salix5:
> > 以前的lflist.conf都是預設從每行讀取2個整數 我是建議genesys分數可以用相同的格式 開頭用其他符號區分就好 例如
> 
> ## Problems on the suggested format
> Sometimes an ordinary banlist (0/1/2) AND a credit system should exist on the same banlist.
> 
> https://cdntx.moecube.com/ygopro-genesys/lflist.conf
> 
> Genesys is also an example that all Link & Pendulums are banned, in which `XXXXX 0` would exist.
> 
> ## Design of this PR
> This system is not only for Genesys, but also for other specific deck build limits, so I added a design for allowing multiple point namespaces.
> 
> Below Rush Duel is an example, for Legend card.
> 
> https://code.moenext.com/mycard/ygopro-rush-duel/-/blob/master/lflist.conf
> 
> Also there are other lflist in communities, something like DL-like server and so-on.
> 
> Btw I just announced this lflist format to them, so no plan to change. Otherwise there would be extra communication costs.

All of your examples are still reading 2 integers from each row,.
Adding the same string to each row will make reading integers harder.
I guess this is what you called "over engineering".

你所有的例子仍然從每一行讀取兩個整數。
在每一行中添加相同的字串會使讀取整數變得更加困難。
我想這就是你所說的「過度設計」。

purerosefallen:
> Adding the same string to each row will make reading integers harder.

It's quite simple, by judging if the 2nd part begins with a `$`, otherwise it's an ordinary entry.

I believe there are requirements that multiple points system exist, so I'm keeping here

purerosefallen:
@mercury233 

How about we support infinite-long texture, for those specific-ruling clients to replace their texture file?

When it comes to 101, we read the thing below that position, which that clients should do them themselves?

purerosefallen:
recently MDPro3 & YGOMobile supported / planning to support this lflist format

purerosefallen:
No interest in maintaining this branch without being reviewed
Planning to be exclusive in https://github.com/purerosefallen/ygopro and other distros such as mdpro3 / ygomobile etc

PR #2935, opened 2025-10-07
Search when enter deck edit

scarletkc:
建议在进入卡组编辑时默认执行一次搜索，以便玩家更好的理解如何组牌。

### DIFF
--- a/gframe/deck_con.cpp
+++ b/gframe/deck_con.cpp
@@ -89,6 +89,7 @@ void DeckBuilder::Initialize() {
 		filterList = &deckManager._lfList.back();
 	}
 	ClearSearch();
+	StartFilter();
 	mouse_pos.set(0, 0);
 	hovered_code = 0;
 	hovered_pos = 0;


purerosefallen:
> 以便玩家更好的理解如何组牌

Could you explain the reason for this change further?

And plz provide a showcase screenshot with before / after

scarletkc:
> > 以便玩家更好的理解如何组牌
> 
> Could you explain the reason for this change further?
> 
> And plz provide a showcase screenshot with before / after

Thanks for the feedback!
The purpose of this change is to help new players who often don’t realize they can click the search bar in the deck edit screen to view all available cards. Because of that, many get stuck not knowing how to start building a deck.
I thought this small tweak could make the UI a bit more intuitive for them.

Here’s a before / after comparison for reference:
before



after
If you think this change isn’t suitable, feel free to close the PR.

yybbwc:
不太行。要做你这个更改，得把重置逻辑也改改，确保搜索栏默认状态一直有卡存在

purerosefallen:
How about to to put `StartFilter` inside `ClearSearch`?

PR #2936, opened 2025-10-11
add default statement SELECT_STMT

salix5:
default statement:
SELECT datas.id, datas.ot, datas.alias, datas.setcode, datas.type, datas.atk, datas.def, datas.level, datas.race, datas.attribute, datas.category,
 texts.name, texts.desc, texts.str1, texts.str2, texts.str3, texts.str4, texts.str5, texts.str6, texts.str7, texts.str8,
 texts.str9, texts.str10, texts.str11, texts.str12, texts.str13, texts.str14, texts.str15, texts.str16 FROM datas INNER JOIN texts ON datas.id=texts.id

https://sqlite.org/lang_select.html#determination_of_input_data_from_clause_processing_
If there is a USING clause then each of the column names specified must exist in the datasets to both the left and right of the join-operator. For each pair of named columns, the expression "lhs.X = rhs.X" is evaluated for each row of the cartesian product as a [boolean expression](https://sqlite.org/lang_expr.html#booleanexpr).

For each pair of columns identified by a USING clause, the column from the right-hand dataset is omitted from the joined dataset. This is the only difference between a USING clause and its equivalent ON constraint. 

----
The header file should explicitly list the column ygopro needs.
頭文件應該明確列出 ygopro 所需的column。

@mercury233 
@purerosefallen 
@Wind2009-Louse
@fallenstardust

### DIFF
--- a/gframe/data_manager.h
+++ b/gframe/data_manager.h
@@ -20,10 +20,6 @@ constexpr int MAX_STRING_ID = 0x7ff;
 constexpr uint32_t MIN_CARD_ID = (uint32_t)(MAX_STRING_ID + 1) >> 4;
 constexpr uint32_t MAX_CARD_ID = 0x0fffffffU;
 
-const char SELECT_STMT[] = "SELECT datas.id, datas.ot, datas.alias, datas.setcode, datas.type, datas.atk, datas.def, datas.level, datas.race, datas.attribute, datas.category,"
-" texts.name, texts.desc, texts.str1, texts.str2, texts.str3, texts.str4, texts.str5, texts.str6, texts.str7, texts.str8,"
-" texts.str9, texts.str10, texts.str11, texts.str12, texts.str13, texts.str14, texts.str15, texts.str16 FROM datas INNER JOIN texts ON datas.id = texts.id";
-
 using CardData = card_data;
 struct CardDataC {
 	uint32_t code{};
--- a/gframe/data_manager.cpp
+++ b/gframe/data_manager.cpp
@@ -6,6 +6,9 @@ namespace ygo {
 
 unsigned char DataManager::scriptBuffer[0x100000] = {};
 DataManager dataManager;
+static const char SELECT_STMT[] = "SELECT datas.id, datas.ot, datas.alias, datas.setcode, datas.type, datas.atk, datas.def, datas.level, datas.race, datas.attribute, datas.category,"
+" texts.name, texts.desc, texts.str1, texts.str2, texts.str3, texts.str4, texts.str5, texts.str6, texts.str7, texts.str8,"
+" texts.str9, texts.str10, texts.str11, texts.str12, texts.str13, texts.str14, texts.str15, texts.str16 FROM datas INNER JOIN texts ON datas.id = texts.id";
 
 DataManager::DataManager() : _datas(32768), _strings(32768) {
 	extra_setcode = { 


purerosefallen:
is JOIN an INNER JOIN or LEFT JOIN?

what if only a half of data exists? (an id only in datas / texts)

salix5:
https://sqlite.org/lang_select.html#special_handling_of_cross_join_
There is no difference between the "INNER JOIN", "JOIN" and "," join operators. They are completely interchangeable in SQLite. The "CROSS JOIN" join operator produces the same result as the "INNER JOIN", "JOIN" and "," operators, but is [handled differently by the query optimizer](https://sqlite.org/optoverview.html#crossjoin) in that it prevents the query optimizer from reordering the tables in the join.

In sqlite
- JOIN
- INNER JOIN
- ,
the same

- CROSS JOIN
produce the same result
prevents the query optimizer from reordering
for special purpose only

purerosefallen:
> There is no difference between the "INNER JOIN", "JOIN" and "," join operators.

I still suggest using common and simple SQL statement not relying on SQLite. This is a common practice as a DBA. (My job has a DBA role in real life) Other SQLite impls, such as SQL.js, Python / Node.js built-in ones, may not support SQLite specific functions.

suggestion:
`INNER JOIN texts ON texts.id = datas.id`

as the upstream project, the other cdb-reading projects would take our repo as reference, so we should use a working but simple, compatible SQL statement.

gframe/data_manager.h
```diff
@@ -19,7 +19,11 @@ namespace ygo {
 constexpr int MAX_STRING_ID = 0x7ff;
 constexpr uint32_t MIN_CARD_ID = (uint32_t)(MAX_STRING_ID + 1) >> 4;
 constexpr uint32_t MAX_CARD_ID = 0x0fffffffU;
-	
+
+const char SELECT_STMT[] = "SELECT datas.id, datas.ot, datas.alias, datas.setcode, datas.type, datas.atk, datas.def, datas.level, datas.race, datas.attribute, datas.category,"
+" texts.name, texts.desc, texts.str1, texts.str2, texts.str3, texts.str4, texts.str5, texts.str6, texts.str7, texts.str8,"
+" texts.str9, texts.str10, texts.str11, texts.str12, texts.str13, texts.str14, texts.str15, texts.str16 FROM datas JOIN texts USING(id)";
```
purerosefallen: Use a common and simple, compatitable SQL
```suggestion
" texts.str9, texts.str10, texts.str11, texts.str12, texts.str13, texts.str14, texts.str15, texts.str16 FROM datas INNER JOIN texts ON texts.id = datas.id";
```

edo9300: FWIW, this is the query I'm using
```
SELECT datas.id,datas.ot,datas.alias,datas.setcode,datas.type,datas.atk,datas.def,datas.level,datas.race,datas.attribute,datas.category,texts.name,texts.desc,texts.str1,texts.str2,texts.str3,texts.str4,texts.str5,texts.str6,texts.str7,texts.str8,texts.str9,texts.str10,texts.str11,texts.str12,texts.str13,texts.str14,texts.str15,texts.str16
FROM datas,texts WHERE texts.id = datas.id ORDER BY texts.id;
```

salix5:
@purerosefallen 
USING(id) is intended

FROM datas INNER JOIN texts ON texts.id = datas.id
dataset:
datas.id, ..., datas.category, texts.id, ...
There are 2 identical `id` column.
The condition in `WHERE` clause has to specify `datas.id`, which is unnecessary because the 2 columns are identical.

FROM datas INNER JOIN texts USING(id)
dataset:
datas.id, ..., datas.category, texts.name, ...
There is only 1 `id` column.
The condition in WHERE clause can be:
WHERE id = 10000


@edo9300 
Yes, this PR is based on observing edopro.
I did not add ORDER BY because:
- it must be put after `WHERE` clause.
There can be other WHERE clause after `SELECT_STMT`.

- ygopro read all rows from the database for now.
The order is not a big problem in this case.

purerosefallen:
> USING(id) is intended

> The condition in WHERE clause has to specify datas.id, which is unnecessary because the 2 columns are identical.

I would strongly recommend against using the USING clause here. While it may look concise, it’s actually less explicit and can easily confuse readers who aren’t deeply familiar with SQLite’s join semantics.

As a DBA, my general principle is to avoid dialect-specific constructs whenever possible — especially those that behave slightly differently across SQL engines. USING implicitly drops one of the joined columns (in this case id), which can create ambiguity and make query logic harder to follow or port to other databases.

The “no need to write datas.id in WHERE clause each time” argument isn’t really a benefit; explicit qualification is preferred in production schemas, as it improves clarity and prevents subtle name conflicts.

To ensure long-term maintainability and readability, I’d suggest sticking to standard ON joins with explicit column references.

salix5:
> As a DBA, my general principle is to avoid dialect-specific constructs whenever possible — especially those that behave slightly differently across SQL engines. USING implicitly drops one of the joined columns (in this case id), which can create ambiguity and make query logic harder to follow or port to other databases.


I appreciate your concerns about SQL portability and explicitness. However, I'd like to clarify a few points:

**Regarding SQL Standard Compliance:**
USING is not "dialect-specific constructs".
USING is part of the SQL:1999 standard and is supported by most major databases including PostgreSQL, MySQL, Oracle, and SQLite (which this project uses).

**About the "Ambiguity" Concern:**
Could you elaborate on what specific ambiguity you're referring to? 
In this case, USING(id) clearly indicates that it will join on the id column, and dropping of the redundant  texts.id column is the intended behavior that simplifies our column indexing.

**About query logic**
datas.id and texts.id are identical row in dataset.
Why would query logic become "harder" without 1 redundant row?
You can still write datas.id if you want.

Before:
datas.id and texts.id refer to 2 identical columns with the same value.

After:
datas.id and texts.id refer to the same column.


Nothing really changes after USING(id).

purerosefallen:
You’re absolutely right that USING is part of the SQL:1999 standard and is supported by most engines — I agree it’s not a SQLite-only construct.
My concern isn’t about “support” per se, but about clarity, predictability, and tooling consistency in real-world usage.

In practice, USING introduces implicit behavior (dropping one side’s column and renaming the join key) that can become confusing in larger queries or when schema definitions evolve.
Explicit ON datas.id = texts.id keeps both the logical relation and the projection behavior visible, which helps with static analyzers, ORMs, and future contributors who might not recall how SQLite flattens join keys.

So the issue isn’t correctness — it’s maintainability and explicitness.
The marginal brevity of USING doesn’t really justify the hidden schema change it introduces, especially in a shared codebase where readability outweighs keystroke savings.

salix5:
> In practice, USING introduces implicit behavior (dropping one side’s column and renaming the join key) that can become confusing in larger queries or when schema definitions evolve.

I believe there may be some misunderstandings here. 
- rename
`USING(id)` doesn't rename anything, it performs an equi-join on the `id` column and returns a single `id` column instead of duplicates. 
This actually makes the result set clearer, not more confusing. 
You can still write datas.id in WHERE clause if you want.

- implicit
USING(id) explicitly drops the id column of the table after INNER JOIN (texts.id in this case).
That is the definition of USING, so I don't understand why it is implicit.


- schema change
Since `id` is the primary key in both tables, it's inherently stable. 
Changing a primary key would require rebuilding the entire application, so this isn't a realistic concern for incremental schema changes. 
Could you provide an example of how `USING(id)` would become problematic when the schema evolves?


> Explicit ON datas.id = texts.id keeps both the logical relation and the projection behavior visible, which helps with static analyzers, ORMs, and future contributors who might not recall how SQLite flattens join keys.

- hidden behavior
I'm not sure what do you mean by "future contributors who might not recall how SQLite flattens join keys".
The complete SQL statement is explicitly defined in the header file as `SELECT_STMT`. 
Any contributor working with this code will see the full query structure, including the `USING(id)` clause. 
There's nothing hidden here.
Even if they don't know "the column from the right-hand dataset is omitted from the joined dataset", it does not matter.
They can still use datas.id or texts.id in WHERE clause if they want.

- maintainability
Could you clarify how `ON datas.id = texts.id` specifically improves maintainability? 
The main difference I see is that `ON` keeps 2 `id` columns with the same value, while `USING(id)` merge them into the same column.

- tool concern
Are you aware of specific static analyzers or ORMs that handle `ON` clauses but have issues with `USING`? 
Both are standard SQL constructs, so tools that understand `JOIN...ON` should handle `JOIN...USING` equally well.

- readability
Can you give an example about how ON increase readability, but USING decrease readability?

purerosefallen:
Thanks for the detailed follow-up — those are good questions, and I’ll try to address them point by point.

1. On schema evolution:
Even if the primary key here is stable, USING can become problematic when additional joins are introduced or when new tables reuse the same column names.
Different SQL engines (and even different client libraries) handle column name resolution and alias precedence differently.
For example, PostgreSQL, MySQL, and SQLite each implement slightly distinct rules for which side’s column name “survives.”
That inconsistency doesn’t affect correctness today, but it makes query portability and schema inspection less predictable later.

2. On maintainability:
Explicit ON datas.id = texts.id improves maintainability not because it changes semantics, but because it makes intent self-documenting.
Reviewers and new contributors can immediately see the join condition and understand which table each column originates from.
USING(id) hides that detail and relies on readers to recall that it merges the columns — which increases cognitive load, especially in multi-join queries.

3. On static analyzers / tools:
Most query analyzers and ORMs that parse SQL (for example, SQLFluff, sqlglot, and some ORM schema diff tools) treat USING differently or with partial support.
Since USING merges columns implicitly, column lineage tracking can become less accurate.
It’s not that they “break,” but that the output of tools like column impact analysis or schema diff reports becomes less deterministic.

4. On readability:
With ON, every join condition explicitly states its relationship — that’s what makes a query readable under review.
With USING, the logic is only obvious if you already know both table schemas and remember how the engine flattens join keys.
In practice, reviewers often have to cross-check the schema to be sure which id is being used, which slows comprehension.

5. On cross-engine consistency:
Even though USING is part of SQL:1999, it isn’t implemented identically everywhere.
Without specific reason or measurable gain, it’s safer and clearer to avoid relying on behavior that might diverge subtly across engines.
The ON form is universally unambiguous and guaranteed to behave the same.

In short — this isn’t about correctness in SQLite, but about clarity, predictability, and long-term consistency.
When the benefit is only syntactic brevity, the explicit ON form tends to scale better for review, tooling, and potential cross-engine use.

gframe/data_manager.h
purerosefallen: > FROM datas,texts WHERE texts.id = datas.id

It's a bit slower than `ON` due to how SQL engine handles

salix5:
> 1. Even if the primary key here is stable, USING can become problematic when additional joins are introduced or when new tables reuse the same column names.
>    Different SQL engines (and even different client libraries) handle column name resolution and alias precedence differently.
>    For example, PostgreSQL, MySQL, and SQLite each implement slightly distinct rules for which side’s column name “survives.”
>    That inconsistency doesn’t affect correctness today, but it makes query portability and schema inspection less predictable later.

Just in case that you did not notice:
id is primary key of both tables.
If 2 primary key columns with identical values are merged into the same column, "which side’s column name survives" does no matter.
I don't have to know where does the merged primary key column come from.

Based on technical documentation, USING(id) behavior with primary keys is actually highly consistent across database systems:

- MySQL, PostgreSQL, SQLite, Oracle: Identical behavior
- All produce single id column, same join semantics

SQLite's USING implementation is identical to other major databases, so the cross-engine consistency concern doesn't apply to our situation.

The claim about "subtle differences across engines" appears to be overstated for primary key joins with USING.



> 4\. With ON, every join condition explicitly states its relationship — that’s what makes a query readable under review.
>     With USING, the logic is only obvious if you already know both table schemas and remember how the engine flattens join keys.
>     In practice, reviewers often have to cross-check the schema to be sure which id is being used, which slows comprehension.

With USING, the logic is obvious if you know the definition of USING and primary key.
USING(id) indicates that id column must be the same.


I don't think USING clause is so complicated that "slows comprehension".
My goal is simple:
- Merge datas, texts into 1 table 
- The table has only 1 primary key column, just like any other SQL table.

If you really think there are some critical points I missed, please give the following examples for your theory: 
- 1 analyze tool that accepts ON clause but rejects USING clause 
- 1 schema change that ON clause can handle while USING clause will fail 
- 1 situation that 2 identical primary key column is better than 1 column

purerosefallen:
> Based on technical documentation, USING(id) behavior with primary keys is actually highly consistent across database systems

which documentation? Yeah you may notice that *investigating documentation* is still costy to ensure USING works here. This is extra work, and risky.

> SQLite's USING implementation is identical to other major databases, so the cross-engine consistency concern doesn't apply to our situation.

as YGOPro, I would expect the thing from `SELECT_STMT` could be simply copied from the source code to other projects reading YGOPro cdb, such as https://code.moenext.com/mycard/ygopro-arena-revive  https://code.moenext.com/mycard/ygopro-cn-database-generator and https://code.moenext.com/mycard/genesys-gen . I don't want to investigate if USING works in SQL.js, Python sqlite3. That's why I say it's risky. I'm unwilling to insert compatitability risk to the project.

> With USING, the logic is obvious if you know the definition of USING and primary key.

It's not obvious because we have to reference the SQLite3 documentation, as well as any DBS and DBMS

> I don't think USING clause is so complicated that "slows comprehension".

Any word that is not common in SQL practice, is complicated, as compatitability & details should be discussed which is costy in project maintainance.

> If you really think there are some critical points I missed, please give the following examples for your theory

you have to give YOUR theory / examples that you think we should use such a keyword, not that I should give counter examples.

salix5:
> > Based on technical documentation, USING(id) behavior with primary keys is actually highly consistent across database systems
> 
> which documentation? Yeah you may notice that _investigating documentation_ is still costy to ensure USING works here. This is extra work, and risky.

https://www.postgresql.org/docs/current/queries-table-expressions.html
Almost all SQL sytem list USING in the same section of ON.


> 
> > SQLite's USING implementation is identical to other major databases, so the cross-engine consistency concern doesn't apply to our situation.
> 
> as YGOPro, I would expect the thing from `SELECT_STMT` could be simply copied from the source code to other projects reading YGOPro cdb, such as https://code.moenext.com/mycard/ygopro-arena-revive https://code.moenext.com/mycard/ygopro-cn-database-generator and https://code.moenext.com/mycard/genesys-gen . I don't want to investigate if USING works in SQL.js, Python sqlite3. That's why I say it's risky. I'm unwilling to insert compatitability risk to the project.

Of course you can copy this statement to any project reading sqlite database.
USING is:
- part of SQL standard 
- explicitly listed in sqlite documentation 

It is not "risky".
It is an ordinary SQL feature.
is there any problem about using sqlite features in different sqlite implementation?


> 
> > With USING, the logic is obvious if you know the definition of USING and primary key.
> 
> It's not obvious because we have to reference the SQLite3 documentation, as well as any DBS and DBMS

How did people know that there is an ON clause?
By reading the document, I suppose.
If people can find ON clause, I think it is not too hard to find USING in the same section.


> 
> > I don't think USING clause is so complicated that "slows comprehension".
> 
> Any word that is not common in SQL practice, is complicated, as compatitability & details should be discussed which is costy in project maintainance.

USING is listed in the same section of ON.
It is as uncommon as ON, I suppose.


> 
> > If you really think there are some critical points I missed, please give the following examples for your theory
> 
> you have to give YOUR theory / examples that you think we should use such a keyword, not that I should give counter examples.

My theory is simple and described above.
Just in case that you didn't see:

My goal is simple:

- Merge datas, texts into 1 table
- The table has only 1 primary key column, just like any other SQL table.

If you really think there are some critical points I missed, please give the following examples for your theory:
- 1 analyze tool that accepts ON clause but rejects USING clause
- 1 schema change that ON clause can handle while USING clause will fail
- 1 situation that 2 identical primary key column is better than 1 column

About examples:
Your comment is full of accusations like:
confusing
risky
reduce readability 
reduce maintainability 
inconsistent in different SQL engines

... which is not understandable or does not exist from my viewpoint.

Of course you have to give at least 1 example for each of the 3 questions.

purerosefallen:
> It is an ordinary SQL feature.

Not a common practice in common SQL use. Have you ever taken a backend dev job or be in another database-related open source project?

> How did people know that there is an ON clause?
> By reading the document, I suppose.

By experiences from project to project.

Have you ever taken a formal course of SQL / DBS design? in those courses `ON foo.id = bar.id` would always be mentioned, being a critial place in SQL practice. However, I hardly see any projects using USING, for not very clear definition, unless referring to the platform-specific DBS documentation.

> USING is listed in the same section of ON.
> It is as uncommon as ON, I suppose.

ON is very common on relational SQL queries, while almost every complex SQL query would include, when joining multiple tables, in case for a 1-to-N or N-to-N query.

> ... which is not understandable or does not exist from my viewpoint.

I believe every DBA would disagree using USING for unclear reason, of course myself being one. http://github.com/koishijs/koishi this is another open source project I've taken place in, in which I contributed many ORM-related PRs.

salix5:
> > It is an ordinary SQL feature.
> 
> Not a common practice in common SQL use. Have you ever taken a backend dev job or be in another database-related open source project?

Just in case that you did not notice:
This is GitHub, bro.
I don't know or care who you are in real life. 

If you are trying to say something like:
"I am a professional IRL, so my opinion is authorative."
I am afraid that it does not work in GitHub.

About the database projects I have participated:
Fluorohydride/ygopro 
class DataManager

salix5/cdb
The database maintained by myself 

mycard/ygopro-database
check if there are bugs when I updated my database 

以防萬一你沒注意到：
兄弟，這是 GitHub。
我不認識你，也不關心你在現實生活中是誰。

如果你想說這樣的話：
“我在現實生活中是專業人士，所以我的觀點是權威的。”
恐怕這在 GitHub 上行不通。

關於我參與過的資料庫專案：
Fluorohydride/ygopro
class DataManager

salix5/cdb
我自己維護的資料庫

mycard/ygopro-database
檢查我更新資料庫時是否有錯誤

> 
> > How did people know that there is an ON clause?
> > By reading the document, I suppose.
> 
> By experiences from project to project.
> 
> Have you ever taken a formal course of SQL / DBS design? in those courses `ON foo.id = bar.id` would always be mentioned, being a critial place in SQL practice. However, I hardly see any projects using USING, for not very clear definition, unless referring to the platform-specific DBS documentation.

Yes.
My teacher always told us that reading the documents by ourselves is an important ability.
Do not trust him or any other authorities, trust what we see in the documents.

If you are trying to say something like:
"My teacher told me using ON clause is better, so my opinion is authorative."
I am afraid that it does not work in GitHub.

是的。
我的老師總是告訴我們，自己閱讀文件是一項重要的能力。
不要相信他或任何其他權威人士，要相信我們在文件中看到的內容。

如果你想說這樣的話：
“我的老師告訴我使用 ON 子句更好，所以我的意見是權威的。”
恐怕這在 GitHub 上行不通。

> 
> > USING is listed in the same section of ON.
> > It is as uncommon as ON, I suppose.
> 
> ON is very common on relational SQL queries, while almost every complex SQL query would include, when joining multiple tables, in case for a 1-to-N or N-to-N query.

It sounds like:
I didn't see USING clause often, so it must be wrong.

聽起來像是：
我很少看到 USING 子句，所以它肯定是錯的。

> 
> > ... which is not understandable or does not exist from my viewpoint.
> 
> I believe every DBA would disagree using USING for unclear reason, of course myself being one. [http://github.com/koishijs/koishi](https://github.com/koishijs/koishi) this is another open source project I've taken place in, in which I contributed many ORM-related PRs.

Does your opinion represent for all DBA in China?
I suppose that it would be a very large number.

By the way , you still didn't give any simple examples.
So I have to ask again.

My goal is simple:

- Merge datas, texts into 1 table
- The table has only 1 primary key column, just like any other SQL table.

If you really think there are some critical points I missed, please give the following examples for your theory:

- 1 analyze tool that accepts ON clause but rejects USING clause
- 1 schema change that ON clause can handle while USING clause will fail
- 1 situation that 2 identical primary key column is better than 1 column

你的觀點能代表中國所有 DBA 的觀點嗎？
我想這應該是相當龐大的人數。

順便說一句，你仍然沒有給出任何簡單的例子。
所以我不得不再問一次。

我的目標很簡單：

- 將資料和文字合併到一張表中
- 該表只有一個主鍵列，就像其他 SQL 表一樣。

如果你真的認為我遺漏了一些關鍵點，請提供以下例子來支持你的理論：

- 一個分析工具接受 ON 子句但拒絕 USING 子句
- 一個 ON 子句可以處理但 USING 子句會失敗的schema更改
- 一個情境，使得兩個相同的主鍵比一個主鍵列更好

salix5:
Use ON now.
Can ones merge this PR?

@mercury233 
@purerosefallen 
@Wind2009-Louse
@fallenstardust

salix5:
For the record:
https://github.com/salix5/ygopro/commit/05d799beade36e50f2e57615eb10b064cc7e7429

PR #2942, opened 2025-11-02
Use miniaudio 0.11.23

purerosefallen:
It worked fine on macOS, but other platforms to be tested


mercury233:
#2833 is fixed in miniaudio 0.11.23

PR #2951, opened 2025-12-28
Fix size calculation for STOC_HS_PlayerEnter

salix5:
Fix the problem in #2551 
Now there is no padding at the end of `STOC_HS_PlayerEnter`.
The size is 41 bytes.


purerosefallen:
how about cross-compatitability?
does it work when old client connecting new server or reverse?

salix5:
Yes
ygopro always accept 41 bytes STOC_HS_PlayerEnter since #2551

purerosefallen:
I mean another side.
Can the new client accept 42-byte PlayerEnter, which the old client sends?

salix5:
Yes.
ygopro only checks:
struct size >= 41

PR #2966, opened 2026-02-04
fix PlaySoundEffect

mercury233:
- Avoid using uninitialized `soundName` in edge cases
- `soundPath` is theoretically not long enough


salix5:
#2967
I suggest that we stop fighting against fixed-size buffer if possible.

PR #2967, opened 2026-02-04
Use std::string in PlaySoundEffect

salix5:
I suggest that we stop fighting against fixed-size buffer if possible.

### DIFF
--- a/gframe/sound_manager.cpp
+++ b/gframe/sound_manager.cpp
@@ -215,12 +215,9 @@ void SoundManager::PlaySoundEffect(int sound) {
 		break;
 	}
 	default:
-		break;
+		return;
 	}
-	std::string soundPath{ "./sound/" };
-	soundPath.reserve(256);
-	soundPath.append(soundName);
-	soundPath.append(".wav");
+	std::string soundPath = "./sound/" + soundName + ".wav";
 	SetSoundVolume(mainGame->gameConf.sound_volume);
 #ifdef YGOPRO_USE_MINIAUDIO
 	ma_engine_play_sound(&engineSound, soundPath.c_str(), nullptr);


gframe/sound_manager.cpp
```diff
@@ -88,137 +88,137 @@ void SoundManager::PlaySoundEffect(int sound) {
...
+		soundName = "chatmessage";
 		break;
 	}
 	default:
 		break;
 	}
-	char soundPath[40];
-	mysnprintf(soundPath, "./sound/%s.wav", soundName);
+	char soundPath[256];
+	mysnprintf(soundPath, "./sound/%s.wav", soundName.c_str());
```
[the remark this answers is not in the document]
mercury233: agree

purerosefallen:
is this PR an alt to https://github.com/Fluorohydride/ygopro/pull/2966?

salix5:
Yes.

gframe/sound_manager.cpp
```diff
@@ -88,143 +88,145 @@ void SoundManager::PlaySoundEffect(int sound) {
...
+		soundName = "playerenter";
 		break;
 	}
 	case SOUND_CHAT: {
-		strcpy(soundName, "chatmessage");
+		soundName = "chatmessage";
 		break;
 	}
 	default:
 		break;
```
mercury233:
```suggestion
		return;
```

gframe/sound_manager.cpp
```diff
@@ -88,143 +88,145 @@ void SoundManager::PlaySoundEffect(int sound) {
...
 	}
-	char soundPath[40];
-	mysnprintf(soundPath, "./sound/%s.wav", soundName);
+	std::string soundPath{ "./sound/" };
+	soundPath.reserve(256);
+	soundPath.append(soundName);
+	soundPath.append(".wav");
```
mercury233:
```suggestion
	std::string soundPath = "./sound/" + soundName + ".wav";
```

No further edit will be done to this string, so it shouldn't be `reserve`d.

mercury233 (changes requested):

PR #2977, opened 2026-02-22
remove _IRR_WCHAR_FILESYSTEM and make DataManager::LoadDB accept char*

mercury233:
require https://github.com/mercury233/irrlicht/pull/16

### DIFF
--- a/gframe/CGUIImageButton.cpp
+++ b/gframe/CGUIImageButton.cpp
@@ -102,7 +102,7 @@ void Draw2DImageQuad(video::IVideoDriver* driver, video::ITexture* image, core::
 	driver->setTransform(irr::video::ETS_VIEW, oldViewMat);
 }
 CGUIImageButton* CGUIImageButton::addImageButton(IGUIEnvironment *env, const core::rect<s32>& rectangle, IGUIElement* parent, s32 id) {
-	CGUIImageButton* button = new CGUIImageButton(env, parent ? parent : 0, id, rectangle);
+	CGUIImageButton* button = new CGUIImageButton(env, parent, id, rectangle);
 	button->drop();
 	return button;
 }
--- a/gframe/CGUITTFont.cpp
+++ b/gframe/CGUITTFont.cpp
@@ -322,7 +322,7 @@ bool CGUITTFont::load(const io::path& filename, const u32 size, const bool antia
 	tt_face = face->face;
 
 	// Store font metrics.
-	FT_Set_Pixel_Sizes(tt_face, size, 0);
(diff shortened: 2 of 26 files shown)


mercury233:
D:\a\ygopro\ygopro\irrlicht\source\Irrlicht\CZipReader.cpp(623,25): warning C4477: 'swprintf' : format string '%s' requires an argument of type 'wchar_t *', but variadic argument 1 has type 'const T *' [D:\a\ygopro\ygopro\build\irrlicht.vcxproj]
need https://sourceforge.net/p/irrlicht/code/5109/#diff-6 to fix the swprintf param type mismatch warning

salix5:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-170
Conversions for character types char and wchar_t are specified by using c or C, and single-byte and multi-byte or wide character strings are specified by using s or S, depending on which formatting function is being used. Character and string arguments that are specified by using c and s are interpreted as char and char* by printf family functions, or as wchar_t and wchar_t* by wprintf family functions.

Microsoft-specific:
The Z type character, and the behavior of the c, C, s, and S type characters when they're used with the printf and wprintf functions, are Microsoft extensions. The ISO C standard uses c and s consistently for narrow characters and strings, and C and S for wide characters and strings, in all formatting functions.

--
Microsoft-specific:
%s means wchar_t string in swprintf

Thanks Microsoft.

salix5:
https://github.com/mercury233/irrlicht/blob/00bc783dc0291b7c53e96fe607454c122d6b7be3/source/Irrlicht/CZipReader.cpp#L623
Maybe need something like 
```cpp
swprintf(buf, 64, L"Not enough memory for decompressing %ls", irr::core::stringw(Files[index].FullName).c_str());
```

Edit:
std::wstring does not have char* constructor. 
Maybe using swprintf deserves tortures like this.

mercury233:
> ```
>      D:\a\ygopro\ygopro\irrlicht\source\Irrlicht\CZipReader.cpp(623,25): warning C4477: 'swprintf' : format string '%s' requires an argument of type 'wchar_t *', but variadic argument 1 has type 'const T *' [D:\a\ygopro\ygopro\build\irrlicht.vcxproj]
> ```
> 
> need https://sourceforge.net/p/irrlicht/code/5109/#diff-6 to fix the swprintf param type mismatch warning

fixed in https://github.com/mercury233/irrlicht/pull/21

purerosefallen:
Need a PR to `mycard/ygopro:server`

purerosefallen:
Good on Linux (tested client, server, server zip)

PR #2979, opened 2026-02-23
update irrlicht clipboard

mercury233:
Our current clipboard-related implementation uses a modified version of Irrlicht, where the parameter type is `wchar_t` instead of the original `char`. This no longer seems necessary.

see https://github.com/mercury233/irrlicht/pull/19

> irrlicht: mercury233/irrlicht@patch-clipboard


purerosefallen:
if the irrlicht PR is intended to be merged, then don't change GitHub Actions file here.

PR #2980, opened 2026-02-25
upgrade Windows x64 Build from SSE2 to AVX2

mercury233:
### Summary

This PR upgrades the **Windows x64** build target from **SSE2** to **AVX2**.

As of **2026**, AVX2 has been available in consumer CPUs for over a decade (introduced in 2013). According to the latest [Steam Hardware Survey](https://store.steampowered.com/hwsurvey/?platform=pc), **96.84% of CPUs support AVX2**, making it a practical and safe baseline for modern Windows x64 systems.
image

### Why AVX2

AVX2 (Advanced Vector Extensions 2) is a 256-bit SIMD instruction set extension for x86 CPUs. It enables wider vector operations, improved integer throughput, and more efficient parallel data processing, leading to measurable performance gains in math-heavy, media, and compute workloads.

### Platform Scope

* **Windows x64**: Upgraded to AVX2 (this PR)
* **Windows x86 (32-bit)**: Remains on SSE2 for compatibility
* **Linux**: Appears to have moved to AVX already; further optimization is not part of this PR
* **macOS**: x86_64 support is being phased out; ARM NEON support requires separate validation and is not part of this PR

### Rationale

Given near-universal AVX2 availability on Windows x64 hardware and its clear performance advantages, continuing to target SSE2 unnecessarily limits modern systems. This change modernizes our baseline while preserving compatibility where still required.

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -260,9 +260,11 @@ workspace "YGOPro"
 
     filter { "system:windows", "platforms:Win32" }
         architecture "x86"
+        vectorextensions "SSE2"
 
     filter { "system:windows", "platforms:x64" }
         architecture "x86_64"
+        vectorextensions "AVX2"
 
     filter "system:macosx"
         libdirs { "/usr/local/lib" }
@@ -314,7 +316,6 @@ workspace "YGOPro"
     filter "action:vs*"
         cdialect "C11"
         conformancemode "On" 
-        vectorextensions "SSE2"
         buildoptions { "/utf-8" }
         defines { "_CRT_SECURE_NO_WARNINGS" }
 


purerosefallen:
does it mean non-avx2 Windows CI runners can no longer build YGOPro?

note: mycard build servers are Sandy Bridge CPUs, which are not avx2

purerosefallen:
And it would negatively affect Windows ygopro servers.
as some VPS providers, like CloudCone, still provide non-avx2 instances

purerosefallen:
I suggest making it as `--avx2` or `--noavx2`

purerosefallen:
at least plz and a premake option. either positively or negatively

mercury233:
will rework on this with more simd and `--no-simd`

PR #2996, opened 2026-03-05
flex deck edit

purerosefallen:
image
image

> note: type count is not included in this PR. It's only from from my fork


purerosefallen:
No interest in maintaining this branch without being reviewed
Planning to be exclusive in https://github.com/purerosefallen/ygopro

PR #2998, opened 2026-03-06
use get_original_code_rule for recursive alias handle

purerosefallen:
https://github.com/Fluorohydride/ygopro/issues/2997

require https://github.com/Fluorohydride/ygopro-core/pull/822

# Support recursive alias resolution via `get_original_code_rule`

## Background

The `alias` field in the card database serves two distinct purposes:

1. **Alternate artwork / reprint** (`is_alternative` returns `true`): the two card IDs are within ±20 of each other. The aliased card is the original printing; the card with the higher ID is a cosmetic variant. For deck-building and ban-list purposes they are treated as the same card.
2. **"Treated as" name** (`is_alternative` returns `false`): the alias points to a card with a very different ID. The card must be counted against the aliased card's copy limit, but they are distinct cards with distinct names.

The existing `card_data::get_original_code()` resolves exactly one hop: if a card is an alternate artwork (`is_alternative` is true) it returns `alias`, otherwise it returns `code`.

## Problem

A new situation has emerged where a **"treated-as" card** — a card whose rules identity is that of another card (`is_alternative` is false, alias points far away) — itself receives alternate artwork printings. The relationship looks like:

```
C (code=20000001, alias=20000000)   ← alternate artwork of B
  → B (code=20000000, alias=37564765, is_alternative=false)   ← "treated as" A
    → A (code=37564765, original)
```

`get_original_code()` for C returns B (`20000000`), because `is_alternative(C, B)` is true and the resolution stops there. C therefore ends up in a different copy-count bucket from A and B, allowing more copies than the rules permit, and the ban-list icon shown for C is incorrect. The correct behavior is for C to inherit B's rules identity — A (`37564765`).

the below line is for PR override

> ocgcore: Fluorohydride/ygopro-core@patch-coderule-backport

### DIFF
--- a/gframe/deck_con.cpp
+++ b/gframe/deck_con.cpp
@@ -1875,21 +1875,21 @@ void DeckBuilder::pop_side(int seq) {
 	GetHoveredCard();
 }
 bool DeckBuilder::check_limit(code_pointer pointer) {
-	auto limitcode = pointer->second.alias ? pointer->second.alias : pointer->first;
+	auto limitcode = get_original_code_rule(pointer->first, pointer->second.alias, DataManager::CardReader);
 	int limit = 3;
 	auto flit = filterList->content.find(limitcode);
 	if(flit != filterList->content.end())
 		limit = flit->second;
 	for (auto& card : deckManager.current_deck.main) {
-		if (card->first == limitcode || card->second.alias == limitcode)
+		if (get_original_code_rule(card->first, card->second.alias, DataManager::CardReader) == limitcode)
 			limit--;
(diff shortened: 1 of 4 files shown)


purerosefallen:
note: build fails because ocgcore submodule ref desync

mercury233:
can you merge master and add instruction in the PR desc as https://github.com/Fluorohydride/ygopro/pull/2999 so we can test the build result

purerosefallen:
https://github.com/Fluorohydride/ygopro/pull/3003

PR #3046, opened 2026-04-09
use shared query_buffer

mercury233:


purerosefallen:
requires an adapt to https://github.com/mycard/ygopro/tree/server

salix5 (commented):
They should use different buffers.
But vectors can be changed to std array.

PR #3052, opened 2026-04-14
update premake define

mercury233:
* Remove the default library path `/usr/local/lib` on macOS
  This is a system default path and should be automatically included by the linker.

* Do not disable `_FORTIFY_SOURCE` on Linux
  `_FORTIFY_SOURCE` helps prevent buffer overflows in functions like `strcpy`.

* Do not suppress the multicharacter constant warning
  `-Wmultichar` warns about multi-character constants enclosed in single quotes.

* Do not suppress the format security warning
  `-Wformat-security` warns about non-literal format strings.
  This pattern is actually used in the codebase:
  https://github.com/Fluorohydride/ygopro/blob/37314171723867a139be8a6b5f91ce7d66c53d10/gframe/duelclient.cpp#L1530

* Drop support for libevent 2.0.x, only support version 2.1.12
  There is no longer a need to maintain compatibility with older versions, and Windows XP support has already been discontinued.

* Stop copying Windows header files via `prebuildcommands`, require users to copy them manually once
  Performing such actions in `prebuildcommands` seems unsafe.

* Use the `event-config.h` generated by CMake instead of the conservative version bundled with libevent
  This resolves issues such as missing includes like `stdint.h`.


mercury233:
```
EVENT__HAVE_STDINT_H : added, avail since VS2010
EVENT__HAVE_UINT64_T : added, in stdint.h
EVENT__HAVE_UINT32_T : added, in stdint.h
EVENT__HAVE_UINT16_T : added, in stdint.h
EVENT__HAVE_UINT8_T : added, in stdint.h
EVENT__HAVE_INTTYPES_H : added, not used
EVENT__HAVE_UINTPTR_T : added

EVENT__SIZEOF_OFF_T : added
EVENT__HAVE__GMTIME64_S : added
EVENT__HAVE__GMTIME64 : added, not used
EVENT__HAVE___func__ : added
EVENT__HAVE___FUNCTION__  : added, not used
EVENT__inline : removed, `inline` is avail in modern MSVC
EVENT__HAVE_STRTOLL : added, avail since VS2015

EVENT__HAVE_ERRNO_H : added, not used in Windows
EVENT__socklen_t=socklen_t : added, not used in Windows
EVENT__HAVE_STRUCT_IN6_ADDR_S6_ADDR32 : removed, not used in libevent itself
EVENT__HAVE_STRUCT_IN6_ADDR_S6_ADDR16 : removed, not used in libevent itself
EVENT__HAVE_DECL_CTL_KERN : added, not used without EVENT__HAVE_SYSCTL
EVENT__HAVE_DECL_KERN_ARND : added, not used without EVENT__HAVE_SYSCTL
EVENT__HAVE_STRINGS_H : removed, not used in libevent itself
EVENT__TIME_WITH_SYS_TIME : removed, not used in libevent itself
EVENT_DNS_USE_FTIME_FOR_ID_: removed, not used in libevent itself
EVENT__DNS_USE_FTIME_FOR_ID : added, not used in libevent itself
EVENT__HAVE_STRUCT_SOCKADDR_STORAGE_SS_FAMILY : added, not used
EVENT__HAVE_STRUCT_LINGER : added, only used in test
EVENT__HAVE_PUTENV : added, only used in test
EVENT__HAVE_UMASK : removed, only used in test
EVENT__SIZEOF_SSIZE_T : added, not used
EVENT__SIZEOF_SOCKLEN_T : added, not used
EVENT__size_t : added, not used
```

PR #3059, opened 2026-04-19
Prevent overwriting existing system/victory strings in ReadStringConfLine

salix5:
ygopro puts a lot of format strings in system strings,
These strings should not be overwritten by expansion files.

### DIFF
--- a/gframe/data_manager.cpp
+++ b/gframe/data_manager.cpp
@@ -191,7 +191,8 @@ void DataManager::ReadStringConfLine(const char* linebuf) {
 		if (std::sscanf(&linebuf[7], "%d %240[^\n]", &value, strbuf) != 2)
 			return;
 		BufferIO::DecodeUTF8(strbuf, strBuffer);
-		_sysStrings[value] = strBuffer;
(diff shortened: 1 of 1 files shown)


gframe/data_manager.cpp
```diff
@@ -183,20 +183,20 @@ void DataManager::ReadStringConfLine(const char* linebuf) {
 	if(linebuf[0] != '!')
 		return;
 	char strbuf[TEXT_LINE_SIZE]{};
-	int value{};
+	unsigned int value{};
```
mercury233: why not uint32_t

mercury233 (approved):

mercury233:
translation expansion will need to be refactored, if there is any

salix5:
ygopro已經將大量的格式字串放在string.conf的system字串
在這個架構下
如果從expansions的conf檔案讀取system字串
同樣也是張開雙手歡迎任何惡意的格式字串
C++函數使用的system字串只能從strings.conf讀取
其他語言翻譯也必須直接修改這個檔案

PR #2727, opened 2025-03-15
use miniaudio

purerosefallen:
reopens https://github.com/Fluorohydride/ygopro/pull/2681

PR #3064, opened 2026-04-23
remove irrklang

mercury233:
[irrklang](https://www.ambiera.com/irrklang/downloads.html) isn't open for downloading now.

miniaudio was added about [one year ago](https://github.com/Fluorohydride/ygopro/pull/2727), and works well.

So I think we can remove the support for irrklang now.

Another question is, should we keep the audio lib select mechanism? If somebody is planning to support another audio lib, it will be useful.


purerosefallen:
*I have said to remove irrklang when introducing miniaudio*

PR #3071, opened 2026-04-27
Remove CTOS_DeckData  and use uint32_t array in deck loading functions

salix5:
# Problem
struct CTOS_DeckData is unnecessary, and it only make the deck loading more complicated.

# Solution
CTOS_UPDATE_DECK
Now it use an ordinary uint32_t array.

The format remains unchanged:
uint32_t mainc, uint32_t sidec, uint32_t[mainc+sidec]

### DIFF
--- a/gframe/deck_manager.cpp
+++ b/gframe/deck_manager.cpp
@@ -147,11 +147,11 @@ uint32_t DeckManager::CheckDeck(const Deck& deck, unsigned int lfhash, size_t ru
 	}
 	return 0;
 }
-uint32_t DeckManager::LoadDeck(Deck& deck, uint32_t dbuf[], int mainc, int sidec, bool is_packlist) {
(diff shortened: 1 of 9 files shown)


mercury233 (approved):

purerosefallen:
is it accepting the data with only that deck long?
eg. only 4*(1+1+40+15+15) bytes

salix5:
> is it accepting the data with only that deck long? eg. only 4*(1+1+40+15+15) bytes

The structure remains unchanged.
Adding the struct is a bad idea, and it only makes the structure more complicated.

The structure is just a uint32_t array

uint_32 data[]
data[0]: mainc, size of (main + extra)
data[1]: sidec, size of side
mainc slots: card id of (main + extra)
sidec slots: card id of side

修改後跟以前一樣都是陣列
只會讀取1+1+mainc+sidec格
後面可以有多餘的資料，只是ygopro不會處理

PR #3083, opened 2026-05-08
Fix size check in STOC_REPLAY and replay read, write

salix5:
The original check will accept `SIZE_MAX - 1`.


salix5:
@mercury233 
By the way, Download NASM failed again.

mercury233:
> @mercury233 By the way, Download NASM failed again.

It is not downloading, it seems to be something wrong with github actions runner, can crash at any step without error message.

PR #3090, opened 2026-05-18
2026.5 tcg banlist

salix5:
Created by Claude Sonnet 4.5
The list is more than 1024 characters long, and it is divided into 2 lines now.
```
2026.5 TCG.txt
It is the previous yugioh TCG banlist.
The format is
${id} ${limit number} --${English name}
The order is exactly the same as the previous banlist on the official website.

name_table_en
the [cid, name] pairs

cid_table
the [id, cid] pairs

https://www.yugioh-card.com/en/limited/list_2026-05-18/
This is the new banlist.
Please update the banlist in the following order
- Find the rows with Remark
- Find cid in name_table_en by its English name
- Find id in cid_table by its cid
- Insert or remove the cards, so that the order stays exactly the same as the new banlist
- The English name in each line should be the name in name_table_en
```


purerosefallen:
*why don't you use Claude to make a tool generating banlist from that site*

mercury233 (approved):

salix5:
> _why don't you use Claude to make a tool generating banlist from that site_

I suppose that I was asking for a code review.

PR #3093, opened 2026-05-22
[WIP] refactor network queue

mercury233:
Changed event handling from direct processing on the network thread to enqueueing events for processing by the main thread, eliminating a large amount of manual locking.


mercury233:
require https://github.com/Fluorohydride/ygopro/pull/3104 to work properly

PR #3097, opened 2026-05-28
add webp

mercury233:
Support WebP-format card images.

TODO:

* Discuss the practical value, since WebP loading is more than 4× slower than JPEG.
* Add WebP support to Irrlicht, so background images and other assets can also use WebP.
* Add options such as `prefer_webp` to avoid repeatedly attempting nonexistent paths when loading WebP-format card images.
* ~~Improve the build system.~~ Done


purerosefallen:
suggest to do after https://github.com/Fluorohydride/ygopro/pull/3099

PR #3106, opened 2026-06-03
fix: prevent crash when multiple global effects trigger at the same time

IMJoyJ:
When multiple global effects (registered via Duel.RegisterEffect with Effect.GlobalEffect) exist at the same event timing, the engine sends MSG_SELECT_CHAIN to the client. The handler card of global effects is temp_card, which has invalid location info (controler=0xFF, location=0xFF). This causes GetCard() to return nullptr, leading to null pointer dereference crash.

Fix by creating a temporary ClientCard via limbo_temp (existing mechanism used by MSG_SELECT_CARD for location=0 cards) when GetCard returns null.

Affected message handlers:
- MSG_SELECT_BATTLECMD
- MSG_SELECT_IDLECMD
- MSG_SELECT_CHAIN

### DIFF
--- a/gframe/client_field.cpp
+++ b/gframe/client_field.cpp
@@ -52,6 +52,10 @@ ClientField::~ClientField() {
 		delete card;
 	}
 	overlay_cards.clear();
+	for (auto& card : limbo_temp) {
+		delete card;
+	}
+	limbo_temp.clear();
 }
 void ClientField::Clear() {
 	for(int i = 0; i < 2; ++i) {
@@ -93,6 +97,10 @@ void ClientField::Clear() {
 		delete card;
 	}
 	overlay_cards.clear();
+	for (auto& card : limbo_temp) {
+		delete card;
+	}
+	limbo_temp.clear();
 	extra_p_count[0] = 0;
 	extra_p_count[1] = 0;
 	player_desc_hints[0].clear();
--- a/gframe/client_field.h
+++ b/gframe/client_field.h
@@ -33,6 +33,7 @@ public:
 	std::vector<ClientCard*> remove[2];
 	std::vector<ClientCard*> extra[2];
 	std::set<ClientCard*> overlay_cards;
+	std::vector<ClientCard*> limbo_temp;
 
 	std::vector<ClientCard*> summonable_cards;
 	std::vector<ClientCard*> spsummonable_cards;
--- a/gframe/duelclient.cpp
+++ b/gframe/duelclient.cpp
@@ -1343,6 +1343,10 @@ bool DuelClient::ClientAnalyze(unsigned char* msg, size_t len) {
 			seq = BufferIO::Read<uint8_t>(pbuf);
 			desc = BufferIO::Read<int32_t>(pbuf);
 			pcard = mainGame->dField.GetCard(con, loc, seq);
+			if(!pcard) {
+				pcard = new ClientCard();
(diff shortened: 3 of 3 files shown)


mercury233:
how to reproduce?

IMJoyJ:
```
--無の煉獄(MODIFIED)
local s,id,o=GetID()
function s.initial_effect(c)
	--Activate
	local e1=Effect.CreateEffect(c)
	e1:SetCategory(CATEGORY_DRAW)
	e1:SetProperty(EFFECT_FLAG_PLAYER_TARGET)
	e1:SetType(EFFECT_TYPE_ACTIVATE)
	e1:SetCode(EVENT_FREE_CHAIN)
	e1:SetCondition(s.condition)
	e1:SetTarget(s.target)
	e1:SetOperation(s.activate)
	c:RegisterEffect(e1)
end
function s.condition(e,tp,eg,ep,ev,re,r,rp)
	return Duel.GetFieldGroupCount(tp,LOCATION_HAND,0)>2
end
function s.target(e,tp,eg,ep,ev,re,r,rp,chk)
	if chk==0 then return Duel.IsPlayerCanDraw(tp,1) end
	Duel.SetTargetPlayer(tp)
	Duel.SetTargetParam(1)
	Duel.SetOperationInfo(0,CATEGORY_DRAW,nil,0,tp,1)
end
function s.activate(e,tp,eg,ep,ev,re,r,rp)
	local p,d=Duel.GetChainInfo(0,CHAININFO_TARGET_PLAYER,CHAININFO_TARGET_PARAM)
	Duel.Draw(p,d,REASON_EFFECT)
	for i=1,3 do --multi global effects
		local ge=Effect.GlobalEffect()
		ge:SetType(EFFECT_TYPE_FIELD+EFFECT_TYPE_CONTINUOUS)
		ge:SetCode(EVENT_PHASE+PHASE_END)
		ge:SetCountLimit(1)
		ge:SetReset(RESET_PHASE+PHASE_END)
		ge:SetLabel(i)
		ge:SetOperation(s.disop)
		Duel.RegisterEffect(ge,p)
	end
end
function s.disop(e,tp,eg,ep,ev,re,r,rp)
	Duel.Hint(HINT_CARD,0,id)
	local i=e:GetLabel()
	if i==1 then
		local g=Duel.GetFieldGroup(e:GetOwnerPlayer(),LOCATION_HAND,0)
		Duel.SendtoGrave(g,REASON_EFFECT+REASON_DISCARD)
	end
end
```

I must apologize for bothering first.

2025-08-24 3caca7504 remove Effect.GlobalEffect
https://github.com/Fluorohydride/ygopro-scripts/commit/3caca7504
https://github.com/Fluorohydride/ygopro-scripts/pull/3006

So there's not any OCG cards in ygopro-scripts using Effect.GlobalEffect() now.
But I still consider it is a significant bug in the client.

# Root Cause
When GetCard() returns nullptr (because the effect handler is temp_card with invalid location 0xFF/0xFF), duelclient.cpp dereferences it immediately — before checking the EDESC_OPERATION flag.

The crash occurs because cn_count >= 2 forces the engine into PROCESSOR_SELECT_CHAIN, which sends MSG_SELECT_CHAIN containing temp_card's invalid location info to the client.

# Why it matters
While OCG scripts no longer use Effect.GlobalEffect(), custom scripts / mods still might do for now.
The client should not crash on any valid engine message — this is logically a missing null-check, not really a script error.

PR #3112, opened 2026-06-16
use std::array for query buffers

salix5:
## update WriteUpdateData parameters 
Remove unnecessary reference parameters.

## use std::array for query buffers in SingleDuel and TagDuel 
16K buffer should be allocated on stack.

## About GetPosition
ocgcore packs 4 bytes into 1 uint32, and the result of `get_infos` is an array of int32/uint32.
Reading the bytes in `info` directly is wrong.

get_infos
Header:
uint32_t clen
length of this card

uint32_t flag
query flag

Content:
Array of uint32_t or int32_t

The flag `QUERY_CODE`. `QUERY_POSITION` is always set in WriteUpdateData
So the first 4 uint32 is always:
Header:
clen
flag

Content:
uint32_t code 
the result of QUERY_CODE

uint32_t info 
the result of QUERY_POSITION
from `get_info_location`

After reading clen, qbuf is at the beginning of flag.
GetPosition always skip the next 2 uint32 (flag, code) and read info directly.

### DIFF
--- a/gframe/network.h
+++ b/gframe/network.h
@@ -197,9 +197,9 @@ struct DuelPlayer {
 	bufferevent* bev{};
 };
 
-inline unsigned int GetPosition(unsigned char* qbuf) {
+inline unsigned int GetPosition(unsigned char* qbuf, size_t offset) {
 	unsigned int info = 0;
-	std::memcpy(&info, qbuf + 2 * sizeof(uint32_t), sizeof info);	// skip flag, code
+	std::memcpy(&info, qbuf + offset, sizeof info);
 	return info >> 24;
 }
 
--- a/gframe/network.h
+++ b/gframe/network.h
@@ -197,9 +197,9 @@ struct DuelPlayer {
 	bufferevent* bev{};
 };
 
-inline unsigned int GetPosition(unsigned char* qbuf, size_t offset) {
+inline unsigned int GetPosition(unsigned char* qbuf) {
 	unsigned int info = 0;
-	std::memcpy(&info, qbuf + offset, sizeof info);
+	std::memcpy(&info, qbuf + 2 * sizeof(uint32_t), sizeof info);	// skip flag, code
 	return info >> 24;
 }
 
--- a/gframe/single_duel.cpp
+++ b/gframe/single_duel.cpp
@@ -1,3 +1,4 @@
+#include <array>
 #include "config.h"
 #include "single_duel.h"
 #include "netserver.h"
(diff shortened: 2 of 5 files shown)


mercury233:
https://github.com/Fluorohydride/ygopro/pull/3046
Why can't a static buffer be used? Does it need to support concurrency or nested calls?

mercury233:
> Reading the bytes in info directly is wrong.

I don't see any issue with it. The data stream contains **mixed** data types, and the value that needs to be read is an 8-bit value, so it is reasonable to read the corresponding byte directly using a `char*`. There is no need to copy it into an `int32` and then perform bit shifting. We also do not consider endianness.

salix5:
> > Reading the bytes in info directly is wrong.
> 
> I don't see any issue with it. The data stream contains **mixed** data types, and the value that needs to be read is an 8-bit value, so it is reasonable to read the corresponding byte directly using a `char*`. There is no need to copy it into an `int32` and then perform bit shifting. We also do not consider endianness.

ygoprou已經把info打包成uint32
因此必須考慮endianness
相信position一定放在qbuf[11]是因為假定宇宙中只有Little Endian
依賴這種不可靠的假定也只是省略一次shift
完全沒有必要

salix5:
> #3046 Why can't a static buffer be used? Does it need to support concurrency or nested calls?

在一般的user space
在stack上面分配16K的buffer是開銷很低的操作
不會對效能產生任何可見的影響
如果每個函數共用global buffer
從此以後每個函數都要跟全域狀態鬥智鬥勇

我想這個問題應該是反過來才對
實際的問題應該是：
有什麼理由讓函數不要自己分配buffer，而是跟global buffer奮戰嗎？

mercury233:
如果要避免依赖大小端序，最好的办法是在`get_info_location`里不再按int32打包。此外我感觉项目中很多地方并不支持大端序架构，而且可见未来里大概也不会出现需要运行ygopro的大端序架构的计算机。

salix5:
我沒有時間或是興趣重寫get_info_location以及ocgapi了
這個PR是我認為還可以的妥協方案

mercury233:
> 我沒有時間或是興趣重寫get_info_location以及ocgapi了 這個PR是我認為還可以的妥協方案

如 https://github.com/Fluorohydride/ygopro/pull/3111#issuecomment-4728274803 所说，这个版本的GetPosition不如修改之前的

salix5:
GetPosition
維持原狀

mercury233 (approved):

PR #3117, opened 2026-06-19
Prevent opening hand result changes before duel start

purerosefallen:
## Summary
- Reject `0` as an invalid opening hand result.
- Ignore repeated opening hand result packets once a side has already submitted.
- Use the tag duel team bit to resolve which side's hand result is being submitted.

## Notes
This keeps the opening rock-paper-scissors result stable after the first valid submission, while still allowing a new submission round after a draw resets the stored results.

### DIFF
--- a/gframe/single_duel.cpp
+++ b/gframe/single_duel.cpp
@@ -354,9 +354,11 @@ void SingleDuel::StartDuel(DuelPlayer* dp) {
 	duel_stage = DUEL_STAGE_FINGER;
 }
 void SingleDuel::HandResult(DuelPlayer* dp, unsigned char res) {
-	if(res > 3)
+	if(res == 0 || res > 3)
 		return;
 	if(dp->state != CTOS_HAND_RESULT)
+		return;
+	if(hand_result[dp->type])
 		return;
 	hand_result[dp->type] = res;
 	if(hand_result[0] && hand_result[1]) {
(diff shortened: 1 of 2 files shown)


gframe/single_duel.cpp
```diff
@@ -354,10 +354,12 @@ void SingleDuel::StartDuel(DuelPlayer* dp) {
 	duel_stage = DUEL_STAGE_FINGER;
 }
 void SingleDuel::HandResult(DuelPlayer* dp, unsigned char res) {
-	if(res > 3)
+	if(res == 0 || res > 3)
```
salix5:
```suggestion
	if(res < 1 || res > 3)
```
The correct value is 1~3

gframe/tag_duel.cpp
```diff
@@ -320,12 +320,12 @@ void TagDuel::StartDuel(DuelPlayer* dp) {
 	duel_stage = DUEL_STAGE_FINGER;
 }
 void TagDuel::HandResult(DuelPlayer* dp, unsigned char res) {
-	if(res > 3 || dp->state != CTOS_HAND_RESULT)
+	if(res == 0 || res > 3 || dp->state != CTOS_HAND_RESULT)
```
salix5, line 323:
```suggestion
	if(res < 1 || res > 3 || dp->state != CTOS_HAND_RESULT)
```

salix5 (dismissed):

gframe/single_duel.cpp
purerosefallen: it's unsigned char so it would not be less than 0

salix5: 我認為從語意來說
拒絕<min, >max比較好懂

purerosefallen: the word `unsigned` already said it's >=0 for sure. Not changing here

salix5: 行吧

salix5 (approved):

mercury233 (approved):

PR #3122, opened 2026-06-21
Increase size of mc array from 5 to 7, fix client MSG_SHUFFLE_SET_CARD

purerosefallen:
sometimes it may be more than 5 cards because there may be at most 7 monsters now shuffling 6 monsters (5 main + 1 extra zone) would crash


mercury233:
should also change other parts, such as `scriptlib::duel_shuffle_setcard`

purerosefallen:
more to do, no time

PR #3125, opened 2026-06-25
Call member functions directly

salix5:
Member functions of Game should call other member functions directly.


mercury233 (changes requested):
The callback in TraversalDir can't omit it, and the other were kept for consistency

PR #3126, opened 2026-06-26
Move default_index to DeckBuilder

salix5:
default_lflist in system.conf
It is the index of _lfList and lflist combobox, and it should be validated before passing to combobox.

default_index
The validated value. 
Now it is managed by DeckBuilder.

### DIFF
--- a/gframe/deck_con.h
+++ b/gframe/deck_con.h
@@ -84,6 +84,7 @@ public:
 	std::mt19937 rnd;
 
 	const LFList* filterList{};
+	size_t default_index{};
 	std::vector<const CardDataC*> results;
 	wchar_t result_string[8]{};
 	std::vector<std::wstring> expansionPacks;
--- a/gframe/deck_con.h
+++ b/gframe/deck_con.h
@@ -84,7 +84,7 @@ public:
 	std::mt19937 rnd;
 
 	const LFList* filterList{};
-	size_t default_lflist_index{};
+	int default_lflist_index{};
 	std::vector<const CardDataC*> results;
 	wchar_t result_string[8]{};
 	std::vector<std::wstring> expansionPacks;
--- a/gframe/deck_con.h
+++ b/gframe/deck_con.h
@@ -84,7 +84,7 @@ public:
 	std::mt19937 rnd;
 
 	const LFList* filterList{};
-	size_t default_index{};
+	size_t default_lflist_index{};
 	std::vector<const CardDataC*> results;
 	wchar_t result_string[8]{};
 	std::vector<std::wstring> expansionPacks;
--- a/gframe/event_handler.cpp
+++ b/gframe/event_handler.cpp
@@ -1951,10 +1951,18 @@ bool ClientField::OnCommonEvent(const irr::SEvent& event) {
 				break;
 			}
 			case CHECKBOX_LFLIST: {
-				mainGame->gameConf.use_lflist = mainGame->chkLFlist->isChecked() ? 1 : 0;
-				mainGame->cbLFlist->setEnabled(mainGame->gameConf.use_lflist);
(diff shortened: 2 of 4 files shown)


gframe/deck_con.h
```diff
@@ -84,6 +84,7 @@ class DeckBuilder: public irr::IEventReceiver {
 	std::mt19937 rnd;
 
 	const LFList* filterList{};
+	size_t default_index{};
```
mercury233: why not `default_lflist_index`

mercury233 (changes requested):

mercury233 (commented):
_don't think it is a good idea to add coupling_

salix5:
> _don't think it is a good idea to add coupling_

談耦合之前還是應該先談正確性
目前版本直接拿用戶輸入的default_lflist傳給ComboBox
這個做法明顯有問題
irrlicht的文件有提到index必須是正常index或-1
傳入錯誤的index沒有立刻崩潰只是因為底層的實作有檢查
但依賴實作的方法既不可靠 也無必要

ygopro應該要保存驗證後的default_lflist_index
由於default_lflist_index主要是由DeckBuilder使用
保存在DeckBuilder是合理的選項

gframe/event_handler.cpp
```diff
@@ -1965,9 +1973,12 @@ bool ClientField::OnCommonEvent(const irr::SEvent& event) {
...
 			switch(id) {
 			case COMBOBOX_LFLIST: {
-				mainGame->gameConf.default_lflist = mainGame->cbLFlist->getSelected();
-				mainGame->cbHostLFlist->setSelected(mainGame->gameConf.default_lflist);
-				mainGame->deckBuilder.filterList = &deckManager._lfList[mainGame->gameConf.default_lflist];
+				int sel = mainGame->cbLFlist->getSelected();
+				if (sel != -1) {
+					mainGame->deckBuilder.default_lflist_index = mainGame->cbLFlist->getSelected();
+					mainGame->cbHostLFlist->setSelected(sel);
+					mainGame->deckBuilder.filterList = &deckManager._lfList[mainGame->deckBuilder.default_lflist_index];
```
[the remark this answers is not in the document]
salix5: Fixed

mercury233 (changes requested):
At the very least, `gameConf.default_lflist` should be removed as part of this change.

But, I don't think we should proceed with this change, since I don't think `deckBuilder.default_lflist_index` should be added in the first place.

gframe/event_handler.cpp
```diff
@@ -1965,9 +1973,12 @@ bool ClientField::OnCommonEvent(const irr::SEvent& event) {
 		case irr::gui::EGET_COMBO_BOX_CHANGED: {
 			switch(id) {
 			case COMBOBOX_LFLIST: {
-				mainGame->gameConf.default_lflist = mainGame->cbLFlist->getSelected();
-				mainGame->cbHostLFlist->setSelected(mainGame->gameConf.default_lflist);
-				mainGame->deckBuilder.filterList = &deckManager._lfList[mainGame->gameConf.default_lflist];
+				int sel = mainGame->cbLFlist->getSelected();
+				if (sel != -1) {
```
mercury233, line 1977: Currently not necessary. One empty lflist is always be pushed to `deckManager._lfList`, also we never clear  `cbLFlist`, nor `setSelected` to `-1`.

salix5: Writing -1 to index is not acceptable.
I don't  think ygopro should fight against every combobox like:
Does this combobox return -1?
Surely not.
Oh wait...

I suppose that checking selected value of any combobox is a good practice.

mercury233: In practice, only the deck list combo box can be empty. The other combo boxes should never be empty in theory, so it's fine to use them directly.

PR #3131, opened 2026-06-29
Refactor ClientCard memory management 

salix5:
# Problem
```cpp
case MSG_MOVE: {
		unsigned int code = BufferIO::Read<int32_t>(pbuf);
		int pc = mainGame->LocalPlayer(BufferIO::Read<uint8_t>(pbuf));
		unsigned int pl = BufferIO::Read<uint8_t>(pbuf);
		int ps = BufferIO::Read<uint8_t>(pbuf);
		unsigned int pp = BufferIO::Read<uint8_t>(pbuf);
		int cc = mainGame->LocalPlayer(BufferIO::Read<uint8_t>(pbuf));
		unsigned int cl = BufferIO::Read<uint8_t>(pbuf);
		int cs = BufferIO::Read<uint8_t>(pbuf);
		unsigned int cp = BufferIO::Read<uint8_t>(pbuf);
		int reason = BufferIO::Read<int32_t>(pbuf);
		if(!mainGame->dInfo.isReplay || !mainGame->dInfo.isReplaySkiping) {
			if(cl & LOCATION_REMOVED && pl != cl)
				soundManager.PlaySoundEffect(SOUND_BANISHED);
			else if(reason & REASON_DESTROY && pl != cl)
				soundManager.PlaySoundEffect(SOUND_DESTROYED);
		}
		int appear = mainGame->gameConf.quick_animation ? 12 : 20;
		if (pl == 0) {
			ClientCard* pcard = new ClientCard;
			pcard->position = cp;
			pcard->SetCode(code);
			if(!mainGame->dInfo.isReplay || !mainGame->dInfo.isReplaySkiping) {
				mainGame->gMutex.lock();
				mainGame->dField.AddCard(pcard, cc, cl, cs);
				mainGame->gMutex.unlock();
				mainGame->dField.GetCardLocation(pcard, &pcard->curPos, &pcard->curRot, true);
				pcard->curAlpha = 5;
				mainGame->dField.FadeCard(pcard, 255, appear);
				mainGame->WaitFrameSignal(appear);
			} else
				mainGame->dField.AddCard(pcard, cc, cl, cs);
		}
```
```cpp
void ClientField::AddCard(ClientCard* pcard, int controler, int location, int sequence) {
	pcard->controler = controler;
	pcard->location = location;
	pcard->sequence = sequence;
	switch(location) {
	case LOCATION_DECK: {
		if (sequence != 0 || deck[controler].size() == 0) {
			deck[controler].push_back(pcard);
		} else {
			deck[controler].insert(deck[controler].begin(), pcard);
		}
		ResetSequence(deck[controler], true);
		pcard->is_reversed = false;
		pcard->ClearData();
		pcard->ClearTarget();
		SetShowMark(pcard, false);
		break;
	}
	case LOCATION_HAND: {
		hand[controler].push_back(pcard);
		ResetSequence(hand[controler], false);
		break;
	}
	case LOCATION_MZONE: {
		mzone[controler][sequence] = pcard;
		break;
	}
	case LOCATION_SZONE: {
		szone[controler][sequence] = pcard;
		break;
	}
	case LOCATION_GRAVE: {
		grave[controler].push_back(pcard);
		pcard->sequence = (unsigned char)(grave[controler].size() - 1);
		break;
	}
	case LOCATION_REMOVED: {
		remove[controler].push_back(pcard);
		pcard->sequence = (unsigned char)(remove[controler].size() - 1);
		break;
	}
	case LOCATION_EXTRA: {
		if(extra_p_count[controler] == 0 || (pcard->position & POS_FACEUP)) {
			extra[controler].push_back(pcard);
		} else {
			size_t faceup_begin = extra[controler].size() - extra_p_count[controler];
			extra[controler].insert(extra[controler].begin() + faceup_begin, pcard);
		}
		ResetSequence(extra[controler], true);
		if (pcard->position & POS_FACEUP)
			extra_p_count[controler]++;
		break;
	}
	}
}
```
MSG_MOVE, pl = 0
If `cl` is wrong, `dField.AddCard` will fail.
The pointer `pcard` will not be placed to any container, and it is lost after leaving the block.
It will causes memory leak.

# Reason
All `ClientCard` object should be managed by `ClientField`, instead of creating new `ClientCard` object at any place.

# Solution
## ClientField
```cpp
std::vector<std::unique_ptr<ClientCard>> cards_;
```
Now it will keep track of `ClientCard` unique_ptr in the private member `cards_`.

## ClientCard
Now it has a pointer to the `ClientField` object.

@Wind2009-Louse

### DIFF
--- a/gframe/duelclient.cpp
+++ b/gframe/duelclient.cpp
@@ -2604,7 +2604,7 @@ bool DuelClient::ClientAnalyze(unsigned char* msg, size_t len) {
 		}
 		int appear = mainGame->gameConf.quick_animation ? 12 : 20;
 		if (pl == 0) {
-			ClientCard* pcard = new ClientCard;
+			ClientCard* pcard = mainGame->dField.CreateCard();
 			pcard->position = cp;
 			pcard->SetCode(code);
 			if(!mainGame->dInfo.isReplay || !mainGame->dInfo.isReplaySkiping) {
@@ -3820,7 +3820,7 @@ bool DuelClient::ClientAnalyze(unsigned char* msg, size_t len) {
 			}
 		} else {
 			while(mainGame->dField.deck[player].size() < mcount) {
-				ClientCard* ccard = new ClientCard;
+				ClientCard* ccard = mainGame->dField.CreateCard();
 				ccard->controler = player;
 				ccard->location = LOCATION_DECK;
 				ccard->sequence = (unsigned char)mainGame->dField.deck[player].size();
@@ -3835,7 +3835,7 @@ bool DuelClient::ClientAnalyze(unsigned char* msg, size_t len) {
 			}
 		} else {
 			while(mainGame->dField.hand[player].size() < hcount) {
-				ClientCard* ccard = new ClientCard;
+				ClientCard* ccard = mainGame->dField.CreateCard();
 				ccard->controler = player;
 				ccard->location = LOCATION_HAND;
 				ccard->sequence = (unsigned char)mainGame->dField.hand[player].size();
@@ -3850,7 +3850,7 @@ bool DuelClient::ClientAnalyze(unsigned char* msg, size_t len) {
 			}
 		} else {
 			while(mainGame->dField.extra[player].size() < ecount) {
-				ClientCard* ccard = new ClientCard;
+				ClientCard* ccard = mainGame->dField.CreateCard();
 				ccard->controler = player;
 				ccard->location = LOCATION_EXTRA;
 				ccard->sequence = (unsigned char)mainGame->dField.extra[player].size();
@@ -3905,13 +3905,13 @@ bool DuelClient::ClientAnalyze(unsigned char* msg, size_t len) {
 			for(int seq = 0; seq < 7; ++seq) {
 				val = BufferIO::Read<uint8_t>(pbuf);
 				if(val) {
-					ClientCard* ccard = new ClientCard;
+					ClientCard* ccard = mainGame->dField.CreateCard();
 					mainGame->dField.AddCard(ccard, p, LOCATION_MZONE, seq);
 					ccard->position = BufferIO::Read<uint8_t>(pbuf);
 					val = BufferIO::Read<uint8_t>(pbuf);
 					if(val) {
 						for(int xyz = 0; xyz < val; ++xyz) {
-							ClientCard* xcard = new ClientCard;
+							ClientCard* xcard = mainGame->dField.CreateCard();
 							ccard->overlayed.push_back(xcard);
 							mainGame->dField.overlay_cards.insert(xcard);
 							xcard->overlayTarget = ccard;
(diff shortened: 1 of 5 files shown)


gframe/duelclient.cpp
mercury233: change here?

mercury233: and so on

mercury233 (changes requested):

gframe/duelclient.cpp
salix5: Now it uses unique_ptr.

mercury233:
Even after this change, there is still some potential memory-safety issues. In `MSG_TAG_SWAP`, after `DestroyCard` is called, containers such as `selectable_cards` and `display_cards` may still hold references to the destroyed cards.

For example, in a tag duel replay, if the user opens the opponent's Deck view before the first player's turn ends, the end-of-turn tag swap will update the Deck contents. The already-open card list can then display incorrect cards, or potentially reference cards that have been destroyed.

Actually, our support for viewing zones such as the Graveyard during replay is already fragile. We currently avoid stale `selectable_cards` issues mostly by forcibly closing `wCardSelect` before `ClientAnalyze` processes new messages.

salix5:
Better than nothing, I suppose.

mercury233 (dismissed):
My understanding is that the goal of this PR is to reduce the risk of memory leaks, but it doesn't address the existing dangling pointer/use-after-free risks, and may actually increase them. If that's the case, I don't think it should be merged now.

salix5:
> My understanding is that the goal of this PR is to reduce the risk of memory leaks, but it doesn't address the existing dangling pointer/use-after-free risks, and may actually increase them. If that's the case, I don't think it should be merged now.

Can you give any example on how this PR "increase" dangling pointer?
It just replaces delete with a new function.

salix5:
Memory leak is a big problem by itself.

If you are trying to say:
This PR only solves 1 big problem,  instead of solving 2 big problems at the same time, so it is not good enough. 

I am afraid that this is not very convincing.

mercury233:
Memory leaks are probably the least serious issue among the "big problems", especially on the client side. A `ClientCard` is less than 1 KB, and over the course of a 20-minute game, it's unlikely that more than a handful of them would ever be leaked.

I'm not saying memory leaks shouldn't be fixed, nor am I saying it's impossible to fix both memory leaks and UAFs at the same time. My point is that UAFs are much more dangerous, and in many cases, strategies for eliminating memory leaks can be at odds with strategies for preventing UAFs.

For this particular PR, I can't point to a specific place where it increases the risk of UAF. Just as static analysis and unit tests can't catch every bug, I can't prove that it does. My concern is that I don't think it's the right direction, or at least not the direction I would choose to pursue.

salix5:
> My understanding is that the goal of this PR is to reduce the risk of memory leaks, but it doesn't address the existing dangling pointer/use-after-free risks, and may actually increase them. If that's the case, I don't think it should be merged now.

First of all:
This PR does not increase UAF or dangling pointer. 
It just replaces delete with a new function.
If ygopro did use a pointer after freed, it already did that before this PR.
I suppose that it is not the case, and your guess  might be wrong. 

# About UAF
The root cause of UAF is creating and destroying objects at any place without being noticed by other functions. 

The main purpose of this PR:
The construction and destruction of ClientCard objects are managed by the vector of smart pointers in ClientField.

By using an array of smart pointers to keep of all active cards, this PR is actually the first step of fixing  UAF.
At least,  it is possible to  check if a pointer is valid by checking `cards` vector.

salix5:
> My point is that UAFs are much more dangerous, and in many cases, strategies for eliminating memory leaks can be at odds with strategies for preventing UAFs.

Again,  can you give any example on "strategies for eliminating memory leaks can be at odds with strategies for preventing UAFs"?

mercury233:
A quick skim of the code wasn't enough for me to fully understand how this PR works. For the reasons I mentioned earlier, I'm not planning to dig into it in depth for now.

Centralizing ownership/management may be the first step toward fixing UAFs, or it may be the first step toward introducing even more bugs (though I also feel the former is more likely).

> can you give any example

One common approach to fixing memory leaks is to add more `free`/`delete` calls, whereas UAFs are often caused by calling `free` or `delete` at the wrong time or in the wrong place.

salix5:
> One common approach to fixing memory leaks is to add more `free`/`delete` calls, whereas UAFs are often caused by calling `free` or `delete` at the wrong time or in the wrong place.

This PR fix memory leaks by using a vector of smart pointers managed by ClientField.
The vector will be cleared or destroyed when Client Field is cleared or destroyed.

If you really worry about use-after-free,  then don't free any ClientCard object as long as ClientField is alive. 
The cards will only be removed from deck or mzone, but the memory is not freed.
As you said, ClientCard objects are small, and keeping all of them would not be a problem.

I suggest that remove all `delete` from the code, and the memory will be freed only when ClientField is cleared or destroyed.

salix5:
@mercury233 
你認為應該移除delete pcard防止UAF嗎？

mercury233:
我不知道，我还是觉得需要更大的重构

_也许在解决ClientThread等线程的生命周期和并发访问之后我会开始研究这个问题_

salix5:
只將new,  delete 替換成由ClientField 管理的函數
這已經是在不破壞現有行為的前提下
我能想到最好的方法
我想作為重構的起點也已經足夠

gframe/duelclient.cpp
```diff
@@ -2602,7 +2602,7 @@ bool DuelClient::ClientAnalyze(unsigned char* msg, size_t len) {
 		}
 		int appear = mainGame->gameConf.quick_animation ? 12 : 20;
 		if (pl == 0) {
-			ClientCard* pcard = new ClientCard;
+			ClientCard* pcard = mainGame->dField.CreateCard();
 			pcard->position = cp;
 			pcard->SetCode(code);
 			if(!mainGame->dInfo.isReplay || !mainGame->dInfo.isReplaySkiping) {
```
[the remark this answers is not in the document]
salix5: One piece at a time, bro.

salix5:
This PR has only one simple purpose:
ClientCard objects created during execution and held by ClientField must be completely reclaimed no later than the time ClientField is destructed or cleared.

This requirement can be easily satisfied if modern C++ memory management methods such as smart pointers are used.

I think this is also the minimum requirement for a C++ project.

--
這個PR只有一個單純的目的：
執行過程中產生且由ClientField保管的ClientCard 物件，最晚在ClientField 解構或清除時必須全部回收。

如果採用smart pointers 等現代C++的記憶體管理，這是能夠輕易滿足的要求。
我想這也是對於C++專案最低限度的要求。

PR #3099, opened 2026-05-29
Refactor premake dependency handling

mercury233:
- Introduce metadata tables and loops to eliminate repeated per-library option declarations and processing code
- Add `--build-all` as a convenience option to build all dependencies from source at once
- **Remove long-deprecated `--use-miniaudio` and `--no-use-miniaudio` options**
- ~~Rename the sqlite subproject from `sqlite3` to `sqlite`~~
- Show a warning when a dependency `include` or `lib` directory is not found
- Make Premake warnings appear in the GitHub Actions annotations
- Update comments

**Bug fixes:**
- `*_INCLUDE_DIR` variables were not converted to absolute paths before being used in `gframe/premake5.lua`, causing relative paths supplied on the command line to be resolved relative to `gframe` instead of the project root
- `no-build-*` flags were checked after `build-*`, allowing an environment variable (e.g. `BUILD_FOO=1`) to silently override an explicit `--no-build-foo` CLI flag
- `GetParam` used Lua's `or` instead of a nil check, causing a falsy `_OPTIONS` value to incorrectly fall back to the environment variable

PR #3140, opened 2026-07-11
define _WIN32_WINNT

mercury233:
Both `WINVER` and `_WIN32_WINNT` should be defined.


purerosefallen:
Is it conflicting with https://github.com/Fluorohydride/ygopro/pull/3099 ?

mercury233:
> Is it conflicting with #3099 ?

???

PR #3144, opened 2026-07-17
update & fix duel timer

mercury233:
- Move ownership of the duel timer from `DuelMode` to `NetServer`.
  - Ready for removing the libevent dependency from `single_duel.cpp` and `tag_duel.cpp` in the [future](https://github.com/mercury233/ygopro/pull/61).
  - The current architecture is already designed to support only one `DuelMode` instance.
- Fix the incorrect file descriptor passed to `event_new` and remove the redundant `EV_TIMEOUT` flag.
- Remove redundant `event_add` and `event_del` calls.
- Change `time_limit` and `time_elapsed` from `short` to `uint16_t`, allowing them to represent the full range of `HostInfo::time_limit`.

### DIFF
--- a/gframe/netserver.cpp
+++ b/gframe/netserver.cpp
@@ -11,11 +11,16 @@ namespace{
 	std::unordered_map<bufferevent*, DuelPlayer> users{};
 	unsigned short server_port{};
 	event_base* net_evbase{};
-	event* broadcast_ev {};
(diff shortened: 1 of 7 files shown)


purerosefallen:
the server uses time_comp.. and time_backed, so a shadow PR to server is needed

PR #3145, opened 2026-07-18
move `miniaudio.h` to `sound_manager.cpp`

mercury233:
- Move the `miniaudio.h` include from `sound_manager.h` to `sound_manager.cpp` to avoid exposing miniaudio declarations to every translation unit that includes the sound manager header.
- Encapsulate miniaudio state and make resource initialization, failure cleanup, and shutdown explicit and safe.
- Remove the heavyweight `game.h` dependency from `sound_manager.h`, replacing it with the required standard headers and a forward declaration.
- Make `RefreshBGMList()` perform a complete refresh by clearing cached entries before rescanning the BGM directories, although it is currently only called once during initialization.


purerosefallen:
it's doing same thing with https://github.com/Fluorohydride/ygopro/pull/3146 moving a lib header file to that cpp file.

Should make into 1 PR

PR #3154, opened 2026-08-05
use default entry point on Windows

mercury233:
Currently, the project uses `main` as its entry point on all platforms. On Windows, `mainCRTStartup` is forcefuly specified as the actual entry point so that it can invoke `main`.

After this change, Windows will use the default `wWinMainCRTStartup` entry point, which invokes `wWinMain`.

Previously, each command-line argument on non-Windows platforms had a fixed maximum length. After this change, a vector of `wstring` objects manages the arguments’ lifetimes, while another vector stores pointers to them.

### DIFF
--- a/gframe/gframe.cpp
+++ b/gframe/gframe.cpp
@@ -25,19 +25,12 @@ void ClickButton(irr::gui::IGUIElement* btn) {
 	ygo::mainGame->device->postEventFromUser(event);
 }
 
-int main(int argc, char* argv[]) {
-#if defined(_WIN32)
-	std::setlocale(LC_CTYPE, ".UTF-8");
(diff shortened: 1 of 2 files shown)


salix5:
Is there any problem to set the limit of each parameter to 256 unicode characters?

mercury233:
> Is there any problem to set the limit of each parameter to 256 unicode characters?

A 256-character limit is usually sufficient, but exceptions such as long file paths exist.

More importantly, the old paths used different types: a fixed two-dimensional array on non-Windows and `wchar_t**` on Windows. Both happened to work locally, but after introducing `mymain`, the array cannot be passed as `wchar_t**`.

PR #3169, opened 2026-09-07
Fix 0 mouse wheel move in deck builder

purerosefallen:
https://github.com/purerosefallen/ygopro/issues/75

Reported from my fork


mercury233:
- sounds like a irrlicht issue
- if fix in ygopro, better check `event.MouseInput.Wheel != 0` ahead.

PR #2526, opened 2024-04-15
submodule use https

salix5:
@mercury233 @purerosefallen 
`git clone` with HTTPS does not require a SSH key or a GitHub account.

PR #3173, opened 2026-09-15
Change submodule URLs to relative paths

salix5:
Fix #2526 

.gitmodules should be flexible.
SSH or HTTPS is up to the developers. 

@Wind2009-Louse


Wind2009-Louse:
As far as I know, some ygopro projects' ocgcore and script modules do not point to the same owner.

salix5:
> As far as I know, some ygopro projects' ocgcore and script modules do not point to the same owner.

Thank you.

mercury233:
I think it would introduce more potential problems than it would solve. Developers can manually specify the submodule path locally without making persistent changes to `.gitmodules`. Meanwhile, local clones and hosting on platforms other than GitHub could both cause issues with relative URLs.

